PR #13, opened 2022-01-17
Fix broken link in getting-started

hjubb:
I think this fixes the 404 link in the getting started page not pointing correctly to `first-steps.md`


onbjerg:
Thanks! :smile:

PR #20, opened 2022-01-24
Add examples to Cheatcodes Reference, and misc

ZeroEkkusu:
Your latest additions to the book are great, ser.

I wanted to make some parts more friendly to new users by providing examples.

If you like this PR, note there are a few cheatcodes I do not know how to use properly and have not provided examples for:
- **ffi** *(here, for example, I don't quite understand the  input, and why it is like that)*
- **getCode** *(couldn't get it to work)*

Also included: Minor additions to `ds-test.md`

🌸

### DIFF
--- a/src/reference/cheatcodes.md
+++ b/src/reference/cheatcodes.md
@@ -70,76 +70,95 @@ interface CheatCodes {
 This section documents all cheat codes, gotchas, and provides usage examples.
 
 #### `warp`
-
 ```solidity
 function warp(uint256) external;
 ```
 
 Sets `block.timestamp`.
+##### Example
 ```solidity
-vm.warp(1641070800);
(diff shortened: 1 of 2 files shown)


onbjerg (changes requested):
Thanks again! I already refer to them as `cheats.fn` so would you kindly either reflect that in your examples or mine? 😄 

I also think we should probably add it under a header (lvl 5) called "Example"

ZeroEkkusu:
✔️ Changed every `vm` to `cheats`
✔️ Placed examples under `##### Example`
✔️ Added horizontal rules to make easier to navigate
❌ Did not add **ffi** and **getCode** examples

onbjerg:
Thanks for this :smile:

PR #24, opened 2022-01-28
Add Cast Guide and begin `cast` Reference

ZeroEkkusu:
- Add Cast Guide
- Begin `cast` Reference
- misc fixes

### DIFF
--- a/src/reference/cast.md
+++ b/src/reference/cast.md
@@ -1,0 +1,60 @@
+## `cast` Reference
+
+`cast` lets you perform Ethereum RPC calls from the comfort of your command line.
+
+### Cast CLI
+
+This is a complete overview of all the available `cast` subcommands. For detailed descriptions and example usage, see below.
+
+```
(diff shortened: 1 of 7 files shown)


src/reference/cast.md
```diff
@@ -0,0 +1,60 @@
+## `cast` Reference
+
+`cast` lets you perform Ethereum RPC calls from the comfort of your command line.
+
+### Cast CLI
+
+This is a complete overview of all the available `cast` subcommands. For detailed descriptions and example usage, see below.
+
+```
```
onbjerg, line 9: Tests are failing because this code block has no specified language, so it just assumes it is Rust. You can add either `ignore` or `text` and it should fix it :blush:

onbjerg (changes requested):

onbjerg:
Thanks!

ZeroEkkusu:
More soon :smile:

PR #28, opened 2022-01-29
Finish `cast` Reference

ZeroEkkusu:


onbjerg:
Not *quite* done yet - we should try to document the environment variables you can set as well. But looks good!

ZeroEkkusu:
Right! I tried listing them for every command, but it looked too bulky.

We'll improve

PR #30, opened 2022-01-30
Basic deployment guide

metehan:


ZeroEkkusu (changes requested):
### Review
- Fix typos
- Refine wording, e.g.
```text
When you need to send arguements to `constructor` of your smart
contract you need to pass `--constructor-args` there can be multiple args.

> REFINE TO:

Use the `--constructor-args` flag to pass arguments to the constructor.
```
- Standardize writing style, e.g.
```bash
$ forge <subcommand>
output goes here
```

ZeroEkkusu:
### Notes
- Further contributions required down the road, e.g. how to verify on Etherscan

onbjerg (changes requested):
Hey @metehan, I appreciate your contribution! Can you address some of @ZeroEkkusu's comments? :smile:

metehan:
@onbjerg yes of course. I was busy.  I will do.

BDANG:
Was going to contribute something similar, but glad I checked the outstanding PRs.

Might be beyond the scope of the foundry book, but it might be helpful to indicate that a local RPC node (and private keys) can be quickly spun up with `npx hardhat node`

onbjerg:
> Was going to contribute something similar, but glad I checked the outstanding PRs.
> 
> Might be beyond the scope of the foundry book, but it might be helpful to indicate that a local RPC node (and private keys) can be quickly spun up with `npx hardhat node`

We could, but a similar feature is being worked in in https://github.com/gakonst/foundry/pull/174

onbjerg:
Also @metehan do you have an approximate timeline for when you can complete this PR? :smile:

metehan:
@onbjerg I just had time, sorry for the late commit :)

Should we also add how to use hardhat as a local development environment? I was also thinking the same as @BDANG. But I decided to not add because hardhat is not our context.

Adding hardhat related information might be useful for starters. We can add it or we can add a new page to the guides.

ZeroEkkusu (commented):
This is better, I can streamline it some more.

**Other things:** 
- Might be better if we swapped the `eth_feeHistory` error tip with the `EIP-1559 not activated` one

onbjerg (approved):
I think this is good for now, we can always improve on it if we find out something is missing. Thanks for your contribution!

PR #32, opened 2022-02-02
additional notes and use cases for expectEmit and expectRevert

mds1:


onbjerg (approved):
Great addition! Thanks :smile:

PR #35, opened 2022-02-12
Fix assert definitions in ds-test reference

ruvaag:
The [ds-test reference](https://github.com/onbjerg/foundry-book/blob/ec66dbc1284f938cbbdeadf3b1bb0c35f3f5813c/src/reference/ds-test.md) mixes up the definitions of `assertGe` & `assertGt` and also those of `assertLe` & `assertLt`.

Fixed after verification against [this](https://github.com/dapphub/ds-test/blob/0a5da56b0d65960e6a994d2ec8245e6edd38c248/src/test.sol) implementation


onbjerg:
Good catch!

PR #40, opened 2022-02-16
Update Forge Guide summary, fix misc in `deploying.md`

ZeroEkkusu:


src/forge/deploying.md
```diff
@@ -1,66 +1,50 @@
...
-#### Error message: Signature error
- > Make sure your private key is correct.
-
-
-
-
-Currently there is an ongoing effort to develop `forge build` command to make deployment experience better. Process can be seen here: https://github.com/gakonst/foundry/issues/402 
+<br>
+
+> ❓ **Troubleshoot**
```
onbjerg, line 40: I think it's a good idea to have this as normal headings for easy permalinks

src/forge/deploying.md
```diff
@@ -1,66 +1,50 @@
...
 ```
 
 ```sh
-$ forge create --rpc-url http://127.0.0.1:8545 --constructor-args ForgeUSD  --constructor-args FUSD  --constructor-args 1000000000000000000000 --private-key df57...3656e MyToken.sol:MyToken
-:MyToken
-compiling...
-no files changed, compilation skipped.
-Deployer: 0x8626f6940e2eb28930efb4cef49b2d1f2c9c1199
-Deployed to: 0x73511669fd4de447fed18bb79bafeac93ab7f31f
+$ forge create --rpc-url <your_rpc_url> --constructor-args "ForgeUSD" "FUSD" 18 1000000000000000000000 --private-key <your_private_key> src/MyToken.sol:MyToken
```
onbjerg, line 36: I'm not entirely sure if this works - there was some oddity with `--constructor-args` and I know we enabled the option to take multiple arguments, but I'm wondering if it is going to swallow `--private-key` and so on. Have you tested it?

ZeroEkkusu: I've never had an issue with it.

ZeroEkkusu:
Something like this?

onbjerg:
I think so. Maybe remove the emoji - we shouldn't really treat it as a callout, it's more like a quick reference for the most common issues, so it should be easily linkable imo. Wdyt?

ZeroEkkusu:
Good call. This is cleaner.

onbjerg:
Thanks again :smile:

PR #46, opened 2022-02-23
chore: update test output

rootulp:
### Motivation
Fixes https://github.com/onbjerg/foundry-book/issues/14 with the exception of https://github.com/onbjerg/foundry-book/issues/45

### Details

Overall the test output doesn't look that different. A few gas amounts have changed. @onbjerg is there anything else about the test output that needs updating?


onbjerg:
Not really, I think it's going to end up being a continual thing until we have a stable release. Thanks for this!

PR #79, opened 2022-03-02
docs: add forge-std's deployCode

rootulp:
## Description

Add docs for https://github.com/brockelmore/forge-std/blob/master/src/stdlib.sol#L78-L88

## Motivation

Screen Shot 2022-03-02 at 1 17 45 AM

### [reference/cheatcodes](https://onbjerg.github.io/foundry-book/reference/cheatcodes.html#getcode)

Screen Shot 2022-03-02 at 10 59 36 PM

### [forge/writing-tests](https://onbjerg.github.io/foundry-book/forge/writing-tests.html)

Screen Shot 2022-03-02 at 10 59 25 PM

### DIFF
--- a/src/forge/writing-tests.md
+++ b/src/forge/writing-tests.md
@@ -65,5 +65,9 @@ Forge uses the following keywords in tests:
 It is possible to use other testing libraries or roll your own. For example, if you find yourself lacking a special type of assertion, you could extend `ds-test`.
 
 > 📚 **Reference**
-> 
+>
 > See the [`ds-test` Reference](../reference/ds-test.md) for a complete overview of the logging functionality and assertions in `ds-test`.
+
+> 💡 **Tip**
+>
+> Use the [`getCode`](../reference/cheatcodes.md#getcode) cheatcode to deploy contracts with incompatible Solidity versions.
--- a/src/reference/cheatcodes.md
+++ b/src/reference/cheatcodes.md
@@ -460,7 +460,7 @@ If a match is found, then `retdata` is returned from the call.
 Mocked calls are in effect until [`clearMockedCalls`](#clearmockedcalls) is called.
 
 > 💬 **Note**
-> 
+>
 > Calls to mocked addresses may revert if there is no code on the address.
 > This is because Solidity inserts an `extcodesize` check before some contract calls.
 >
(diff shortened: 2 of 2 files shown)


ncitron:
Everything here looks correct to me. I would also probably mention that this cheatcode is very useful for deploying contracts that must be compiled with a different version of solidity than your test contract. That was the main usecase I had in mind when creating the cheatcode.

I'm not sure if this is too in depth for this section, but if you look in the forge README it goes more in depth into the ways that a contract file can be referenced.

From the README:
Fetches bytecode from a contract artifact. The parameter can either be in the form `ContractFile.sol` (if the filename and contract name are the same), `ContractFile.sol:ContractName`, or `./path/to/artifact.json`.

ZeroEkkusu (commented):
Everything Citron said, but also show how to use `getCode`:
```solidity
MyContract myContract = new MyContract(arg1, arg2);

// Let's do the same thing with `getCode`
bytes memory args = abi.encode(arg1, arg2);
bytes memory bytecode = abi.encodePacked(cheats.getCode("MyContract.sol:MyContract"), args);
address anotherAddress;
assembly {
    anotherAddress := create(0, add(bytecode, 0x20), mload(bytecode))
}

assertEq0(address(myContract).code, anotherAddress.code); // [PASS]
```
*Useful for deploying contract with incompatible Solidity versions...
Also `out/MyContract.sol/MyContract.json` or `MyContract.sol` if contract name is the same...
Better use forge-std's `deployCode`..*

ZeroEkkusu:
> I think that may not be the most discoverable. Defer to reviewers on if there's a better location for this.

You may mention it in [`Writing Tests`](https://github.com/onbjerg/foundry-book/blob/master/src/forge/writing-tests.md) at the end of the page.

Use the "Tip" format:
> 💡 **Tip**
> 
> Use the [`getCode`](https://onbjerg.github.io/foundry-book/reference/cheatcodes.html#getcode) cheatcode to deploy contracts with incompatible Solidity versions.

You can see an example of a tip [here](https://onbjerg.github.io/foundry-book/getting-started/first-steps.html).

rootulp:
^oof sorry I accidentally referenced this issue in an unrelated PR.

Thanks for the super helpful comments! I will update docs to include a tip in **Writings Tests** and a code sample for `getCode`!

rootulp:
^ Added @ncitron @ZeroEkkusu as co-authors, I hope that's alright.

onbjerg (approved):
Looks good to me, sorry for the slow turnaround on this one guys 😞 Thanks for the contributions, both in reviewing and the PR!

@all-contributors please add @ncitron for review

onbjerg:
Yooo... @all-contributors please add @ncitron for review 🤔

PR #85, opened 2022-03-06
3 commits!

### LINKED ISSUE #65: Add section on discarding fuzzer inputs
In our fuzzing guide we should make use of `vm.assume`, or at least have a section dedicated to it, as it is the preferred tool to discard fuzzer inputs.

Ref: https://github.com/onbjerg/foundry-book/issues/37 / #60

ZeroEkkusu:
## fix: misc
Fixes. No new info.

## docs(contributing): add `Writing Style`
Inform new contributors about some of the writing standards used throughout the book.
E.g. provide console outputs, use relative links, etc.

## feat: explain essential but non-obvious things
Such as combining subcommands, mainnet forking, shared setups, and many more.

## fix: add misc missing info and remove `Off-topic Guides`, forking
Add the missing cheatcodes examples, cast envs, and the other missing info.
Remove `Off-topic Guides` and forking.
Also, closes #65 if this is enough.


src/forge/working-on-an-existing-project.md
```diff
@@ -9,7 +9,7 @@ $ git clone --recursive https://github.com/brockelmore/forge-std
 $ cd forge-std
 ```
 
-Notice that we clone with the `--recursive` flag. This automatically pulls git submodules (which we use as dependencies) as well in case there are any.
+Notice that we clone with the `--recursive` flag. This automatically pulls git submodules (which we use as dependencies) as well in case there are any. Alternatively, run `forge update` after cloning to fetch dependencies.
```
onbjerg, line 12: It's `forge install` :smile:

src/guides/offtopic.md
```diff
@@ -0,0 +1,51 @@
+## Off Topic Guides
```
onbjerg, line 1: I'm not sure these guides belong in the book for a couple of reasons.

We have to assume some familiarity with certain things. One of the things we already assume familiarity with is Solidity, another the command line. For that reason, we probably shouldn't have a guide on environment variables - at the same time, we probably shouldn't have a guide on `make` either, since it's really up to people if they want to use bash or makefiles or powershell or whatever to write their custom scripts.

src/forge/tests.md
```diff
@@ -46,6 +46,12 @@ Running 1 test for ContractC.json:ContractC
 
 The inverse of the `--match-path` flag is `--no-match-path`.
 
+You can run tests on a forked chain - such as the Ethereum mainnet - by passing a RPC URL via the `--fork-url` flag:
+
+```bash
+$ forge test --fork-url <your_rpc_url>
+```
+
```
onbjerg, line 54: I think we should probably have a more dedicated chapter on fork tests since it changes a couple of things about the test runner, such as the chain ID. In a more dedicated guide we also have room to explain things like forking from a specific block

src/guides/offtopic.md
ZeroEkkusu: k

src/forge/working-on-an-existing-project.md
ZeroEkkusu: both work, will change

src/forge/tests.md
ZeroEkkusu: i agree

onbjerg:
Thanks! Appreciate the quick turnaround (and again, the contributions :smile:)

PR #91, opened 2022-03-09
clarify assume use cases

mds1:
I've seen users confused over when it's appropriate to use `assume`, so wanted to clarify that

### DIFF
--- a/src/reference/cheatcodes.md
+++ b/src/reference/cheatcodes.md
@@ -2,7 +2,7 @@
 
 Cheatcodes give you the ability to alter the state of the EVM, mock data, assert on reverts, and more.
 
-To enable a cheatcode you call designated functions on the cheatcode address: `0x7109709ECfa91a80626fF3989D68f67F5b1DD12D`.
+To enable a cheatcode you call designated functions on the cheatcode address: `0x7109709ECfa91a80626fF3989D68f67F5b1DD12D`. This address can be accessed through the `HEVM_ADDRESS` constant in `DSTest`
 
 ### Cheatcodes Interface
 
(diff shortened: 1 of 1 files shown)


src/reference/cheatcodes.md
```diff
@@ -2,7 +2,7 @@
 
 Cheatcodes give you the ability to alter the state of the EVM, mock data, assert on reverts, and more.
 
-To enable a cheatcode you call designated functions on the cheatcode address: `0x7109709ECfa91a80626fF3989D68f67F5b1DD12D`.
+To enable a cheatcode you call designated functions on the cheatcode address: `0x7109709ECfa91a80626fF3989D68f67F5b1DD12D`. This address can be accessed through the `HEVM_ADDRESS` constant in `DSTest`
```
onbjerg, line 5:
```suggestion
To enable a cheatcode you call designated functions on the cheatcode address: `0x7109709ECfa91a80626fF3989D68f67F5b1DD12D`. This address can be accessed through the `HEVM_ADDRESS` constant in `DSTest`.
```

src/reference/cheatcodes.md
```diff
@@ -606,12 +606,24 @@ function assume(bool) external;
 
 If the boolean expression evaluates to false, discard the current fuzz inputs and start a new fuzz run.
 
+The `assume` cheatcode should mainly be used for very narrow checks. Broad checks will slow down tests as it will take a while to find valid values, and the test may fail if you hit the max number of rejects. You can also [configure](https://github.com/gakonst/foundry/blob/a00c1939becf3b40f1453e330d3d9507a756d610/config/src/lib.rs#L181-L188) `fuzz-max-local-rejects` and `fuzz-max-global-rejects` in your `foundry.toml` file.
```
onbjerg, line 609: Rather than linking to a source file here it would be better if we add these keys to the `foundry.toml` reference and link to them in the book :smile:

onbjerg (changes requested):
Some minor changes! Otherwise great addition :smile:

src/reference/cheatcodes.md
mds1: fixed in 1388ed7

src/reference/cheatcodes.md
mds1: Good call, updated in 1388ed7. Also added a link to the proptest book which has more info on filtering

ZeroEkkusu (changes requested):
We should probably change the fuzzing example with `assume` in [Fuzz Testing](https://onbjerg.github.io/foundry-book/forge/fuzz-testing.html) (at the end of the page) since it is a broad check.

mds1:
In that example, we're only excluding `1e17 / 2^96` = 1e-10% of values, so that check is probably sufficiently narrow. I tested it which confirms there are no errors with the default rejection settings. Even flipping the sign is ok. But using `vm.assume(amount > 100 && amount < 105)` fails, because that check has a very tight arbitrary requirement while the other two are wider ranges with "patterns" that I think proptest can figure out

Though I see how the current example could be defined as a broad check, so maybe we need better terminology? Not sure how to concisely describe the cases yet 🤔

onbjerg:
Let's follow up in another PR, the fuzzer is still being improved so I doubt this is the last time we have to update these docs 😄 Thanks for the PR!

PR #96, opened 2022-03-11
docs: label name param data location = calldata

devtooligan:
label() is missing data location for name param elicits:
Screen Shot 2022-03-11 at 11 36 02 AM


onbjerg:
Thanks 😄 

@all-contributors add @devtooligan for maintenance

PR #97, opened 2022-03-11
Add foundry rust template to references

nikkolasg:


ZeroEkkusu (commented):
Gm @nikkolasg!

I don't think this is the best place to put it (certainly not before the original book content).
Also, might be better to add it to [Awesome Foundry](https://github.com/crisgarner/awesome-foundry), since there's no mention of it there 🧐.

onbjerg:
I appreciate the contribution but I agree this fits better in https://github.com/crisgarner/awesome-foundry :smile:

PR #98, opened 2022-03-12
Update expectEmit reference

odyslam:
- Remove requirement for not checking values in batched events
- Add best practice comment on defining the event ourselves if it's not in-scope.

This is based on what I learned in the recent uniswap v3-periphery migration

### DIFF
--- a/src/reference/cheatcodes.md
+++ b/src/reference/cheatcodes.md
@@ -425,9 +425,14 @@ This cheat code is used to assert that certain logs are emitted on the next call
 2. Emit the event we are supposed to see after the next call.
 3. Perform the call.
 
+If the event is not available in the scope (e.g because we use some Interface or external smart contract), we can define the event ourselves with an identical event signature. 
+
+
 For example:
 
 ```solidity
+event Transfer(address indexed from, address indexed to, uint256 amount);
+
 function testERC20EmitsTransfer() public {
   // Only `src` and `dst` are indexed in ERC20's `Transfer` event,
   // so we only check topic 0 and 1, as well as the data (`amount`).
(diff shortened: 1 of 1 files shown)


ZeroEkkusu (approved):
Thanks! Good additions.

We can change `Transfer`'s  topics' names to `src` and `dst` to be consistent with the docs, as well as use just `emit Transfer` without `MyToken`, but we can do all of that later.

src/reference/cheatcodes.md
```diff
@@ -425,9 +425,15 @@ This cheat code is used to assert that certain logs are emitted on the next call
 2. Emit the event we are supposed to see after the next call.
 3. Perform the call.
 
+If the event is not available in the scope (e.g because we use some Interface or external smart contract), we can define the event ourselves with an identical event signature.
```
onbjerg, line 428:
```suggestion
If the event is not available in the current scope (e.g because we are using an interface, or an external smart contract), we can define the event ourselves with an identical event signature. 
```

src/reference/cheatcodes.md
```diff
@@ -425,9 +425,15 @@ This cheat code is used to assert that certain logs are emitted on the next call
 2. Emit the event we are supposed to see after the next call.
 3. Perform the call.
 
+If the event is not available in the scope (e.g because we use some Interface or external smart contract), we can define the event ourselves with an identical event signature. 
+
+Currently, the cheatcode does not check the origin contract of the event, but simply that it was emitted during that call.
```
onbjerg, line 430:
```suggestion
The cheatcode does not check the origin of the event, but simply that it was emitted during that call.
```

src/reference/cheatcodes.md
```diff
@@ -461,14 +467,15 @@ function testERC20EmitsTransfer() public {
...
 
-We can also assert that multiple events are emitted in a single call. However this currently requires that we only test event emission and not topics or data. For example:
+We can also assert that multiple events are emitted in a single call. 
+For example:
 
 ```solidity
 function testERC20EmitsBatchTransfer() public {
   // We declare multiple expected transfer events
   for (uint256 i = 0; i < users.length; i++) {
     // Each parameter must be set to false for batch event emission tests to work.
```
onbjerg, line 477: This comment is no longer true

src/reference/cheatcodes.md
```diff
@@ -461,14 +467,15 @@ function testERC20EmitsTransfer() public {
 
 <br>
 
-We can also assert that multiple events are emitted in a single call. However this currently requires that we only test event emission and not topics or data. For example:
+We can also assert that multiple events are emitted in a single call. 
+For example:
```
onbjerg, line 471: Consolidate to a single line
```suggestion
We can also assert that multiple events are emitted in a single call. For example:
```

onbjerg (commented):
Thanks Ody :smile: Minor nits otherwise this is good to merge

onbjerg:
@all-contributors add @odyslam for maintenance

PR #105, opened 2022-03-12
Forking mode and `CONTRIBUTING` additions

ZeroEkkusu:


onbjerg:
I've been thinking that we should move the section about gas reports out from the testing chapter, possibly before the "Continous Integration" chapter. Gas reports will probably logically fit in the same chapter as gas snapshots as well, so maybe we should rename the chapter at some point to "Gas tracking" or something. Would you mind moving gas reports out in this PR?

onbjerg:
Thanks! :smile:

PR #108, opened 2022-03-16
zero index event topics

wildmolasses:
cheers!


cruzdanilo:
is this correct? from what i get, `topic0` (selector) is always checked by default

wildmolasses:
Ah crap, yes wrong.

Instead, this line should be changed: https://github.com/onbjerg/foundry-book/blob/master/src/reference/cheatcodes.md?plain=1#L439

PR #114, opened 2022-03-21
Fix checksum for test contract address

rootulp:
### Description
Fix checksum for test contract address. You can verify this checksum by navigating to https://etherscan.io/address/0xb4c79dab8f259c7aee6e5b2aa729821864227e84 and comparing the address (case-sensitive) that Etherscan displays with the address casing in this PR.

### Motivation
Screen Shot 2022-03-20 at 8 12 30 PM


onbjerg:
Good catch, thank you!

PR #117, opened 2022-03-21
fixing route for tutorial after renaming

FredCoen:
@onbjerg hey sry missed the renaming of the path :)


onbjerg:
You should also rename the file itself :smile:

FredCoen:
think i did that. this one you mean? https://github.com/onbjerg/foundry-book/blob/master/src/tutorials/solmate-nft.md

onbjerg:
Ahh, my bad!

PR #120, opened 2022-03-22
Added clarifying info about forge create

nnnnicholas:
Updated the Deploying and Verifying page to clarify part of the CLI parameter and to specify that `forge create` can only deploy one contract at a time.


onbjerg:
Thanks!

PR #121, opened 2022-03-22
Add calldata to label function signature

rootulp:
Fixes

Screen Shot 2022-03-21 at 11 05 47 PM

Note this function signature already includes `calldate` [here](https://github.com/onbjerg/foundry-book/blob/bcc23b30a453692240f7f38491f603064f24daa1/src/reference/cheatcodes.md?plain=1#L90)


onbjerg:
Thanks! :smile:

PR #130, opened 2022-03-22
Update shell-autocompletion to ensure autoload is .zshrc

devanonon:


onbjerg:
@all-contributors add @devanonon for docs

onbjerg:
Not sure if it checks out (don't use zsh myself), but I assume it does, so thanks! :smile:

devanonon:
Thanks @onbjerg! I ran into this issue and found the solution from [GitHub's CLI docs](https://cli.github.com/manual/gh_completion). So, that can act as a second source 😄

Issue #133, opened 2022-03-25
Creating an NFT with Solmate guide would lock funds in contract


abcoathup:
Error in guide: https://onbjerg.github.io/foundry-book/tutorials/solmate-nft.html

Withdraw function uses OpenZeppelin PullPayment but mint function doesn't transfer funds to [escrow](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/76fca3aec8e6ce2caf1c9c9a2c8396fe0882591a/contracts/security/PullPayment.sol#L67)

Funds would be locked in contract.  
Guide should include test to check funds can be withdrawn.

Guide doesn't compile as Nft used for both contract and test contract.

dmfxyz:
Addressed the compiling issues in #137

abcoathup:
Thanks @dmfxyz for fixing the tutorial.  Didn't want someone to copypasta and end up with locked funds.

PR #137, opened 2022-03-29
small fixes to solmate nft tutorial

dmfxyz:
I am writing a Docker tutorial that builds off the awesome solmate tutorial currently in the foundry book. While writing it, I noticed a few minor bugs in the NFT tutorial:

1. Does not explicitly instruct user to install openzeppelin dependencies
2. The "Basic NFT" won't actually compile because it does not implement the `tokenURI` function from Rari's ERC721 abstract contract.

I just made a couple edits to address these small "bugs".

### DIFF
--- a/src/tutorials/solmate-nft.md
+++ b/src/tutorials/solmate-nft.md
@@ -7,7 +7,7 @@ This tutorial walk you through creating an OpenSea compatible NFT with Foundry a
 Start by setting up a Foundry project following the steps outlined in the [Getting started section](../getting-started/installation.html). We will also install Solmate for their ERC721 implementation, as well as some OpenZeppelin utility libraries. Install the dependencies by running the following commands from the root of your project:
 
 ```bash
-forge install Rari-Capital/solmate Openzeppelin/openzeppelin-contracts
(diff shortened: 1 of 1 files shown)


onbjerg (approved):
This looks great - thank you. Any chance I could ask you to fix the remaining issues in #133 as well? If not, no big deal!

onbjerg:
@all-contributors add @dmfxyz for maintenance

dmfxyz:
@onbjerg Yeah, no problem. I have to get it all in a working state for the docker tutorial anyway :)

I think we should add a disclaimer with the tutorials. Something like "This tutorial is for illustrative purposes only and provided on an as-is basis. The tutorial is not audited nor fully tested. No code in this book should be used in a production environment without proper testing and auditing."

Thoughts? Foundry is maintained by a number of leaders in this space, so I could see people coming here and just copy pasting...

dmfxyz:
@onbjerg I've removed the OZ PullPayments (it's overkill for this tutorial) and replaced it with a simple onlyOwner withdrawal method. Added tests for the withdrawal. Also added disclaimer. I think this should address everything in #133 and be good to merge, and then I can build the docker tutorial off of that.

onbjerg:
Thank you!!!

PR #140, opened 2022-03-30
Fix incorrect type on return value of address().code in the 'etch' ch…

DrakeEvans:
…eat code example

Etch cheat code example incorrectly had the return type as address instead of bytes


onbjerg:
Thanks! @all-contributors add @DrakeEvans for maintenance

PR #155, opened 2022-04-01
fix: typos in NFT tutorial

paulrberg:
The main problem is that the `nft-tutorial` hyperlink has an extra parenthesis (`(`), which causes the link to show up like this:

Screen Shot 2022-04-01 at 12 21 45 PM

### DIFF
--- a/src/tutorials/solmate-nft.md
+++ b/src/tutorials/solmate-nft.md
@@ -75,7 +75,9 @@ If successfully deployed, you will see the deploying wallet's address, the contr
 
 Calling functions on your NFT contract is made simple with Cast, Foundry's command-line tool for interacting with smart contracts, sending transactions, and getting chain data. Let's have a look at how we can use it to mint NFTs from our NFT contract.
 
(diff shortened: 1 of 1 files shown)


dmfxyz (commented):
can you replace ` function testFailWithdrawalAsNotOwner()` test with the following? This generates a false positive pass because of the combination of `fail` and `expectRevert`. 

```solidity
function testWithdrawalFailsAsNotOwner() public {
        // Mint an NFT, sending eht to the contract
        Receiver receiver = new Receiver();
        nft.mintTo{value: nft.MINT_PRICE()}(address(receiver));
        // Check that the balance of the contract is correct
        assertEq(address(nft).balance, nft.MINT_PRICE());
        // Confirm that a non-owner cannot withdraw
        vm.expectRevert("Ownable: caller is not the owner");
        vm.startPrank(address(0xd3ad));
        nft.withdrawPayments(payable(address(0xd3ad)));
        vm.stopPrank();
    }
```

I know you didn't write this (I did, actually), but would be good to fix in this PR.

PR #158, opened 2022-04-02
Refactor code snippets to use custom errors

paulrberg:
This PR depends on #157 being merged first.

If you decide not to merge #157, I'd be happy to reapply my changes here.

### DIFF
--- a/src/forge/cheatcodes.md
+++ b/src/forge/cheatcodes.md
@@ -96,15 +96,14 @@ $ forge test -vvvv
 compiling...
 no files changed, compilation skipped.
 Running 2 tests for OwnerUpOnlyTest.json:OwnerUpOnlyTest
-[PASS] testFailIncrementAsNotOwner() (gas: 4030)
+[P[PASS] testFailIncrementAsNotOwner() (gas: 10406)
 Traces:
-
-  [4030] OwnerUpOnlyTest::testFailIncrementAsNotOwner()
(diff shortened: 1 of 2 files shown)


onbjerg (changes requested):
The guide in src/forge/cheatcodes.md is no longer correct. We revert with a custom error, but later on we use `expectRevert`, expecting a revert with a string error

paulrberg:
@onbjerg I fixed the the cheatcodes in the last commit.

I'm not sure though why the total gas limit is higher than the values that you wrote initially .. perhaps I ran the tests with different compiler settings, or the latest foundry CLI has become more expensive in terms of gas costs when running the test?

src/forge/cheatcodes.md
```diff
@@ -95,15 +96,14 @@ $ forge test -vvvv
 compiling...
 no files changed, compilation skipped.
 Running 2 tests for OwnerUpOnlyTest.json:OwnerUpOnlyTest
-[PASS] testFailIncrementAsNotOwner() (gas: 4030)
+[P[PASS] testFailIncrementAsNotOwner() (gas: 10406)
```
onbjerg, line 99:
```suggestion
[PASS] testFailIncrementAsNotOwner() (gas: 10406)
```

onbjerg (changes requested):

PR #162, opened 2022-04-03
expand `expectRevert` on custom errors

wschwab:
Hi!

I was working on writing tests for some custom errors (without parameters) in Foundry, and ended up getting stuck on a point here - it's vanilla Solidity, but I figured also not necessarily common knowledge (<= me coping with not knowing), so I figured I'd add in a little paragraph here in case someone else gets stuck on it.

Feel free to ask for changes or the like - I just put it together in a way that made sense to me.


wschwab:
I actually realized this edit isn't very clear, and mixes the different signatures of `expectRevert` together - I'm going to close this PR and try to get back to this when I think I can try to get across what got me stuck more clearly

PR #174, opened 2022-04-06
Add caveat about order of calls in "expectEmit"

paulrberg:
I spent one hour bashing my head against the wall trying to figure out why my test failed - the reason was that `expectEmit` was eavesdropping on a call I made to read a storage variable.


onbjerg:
This should not be the case. `expectEmit` just collects the next log, not much else.

paulrberg:
> This should not be the case. expectEmit just collects the next log, not much else.

Then I may have found a bug. Do you want me to open an issue with steps to reproduce?

onbjerg:
Yes please

paulrberg:
Done.

https://github.com/gakonst/foundry/issues/1214

PR #184, opened 2022-04-08
update `cast` reference

zk-tarts:
I plan to add examples for the undocumented `cast` commands.  
Right now I've got these ones done.
- [x] `--from-bin`
- [x] `--from-fix`
- [x] `--to-int256`
- [x] `--to-unit`
- [x] `4byte-event`
- [x] `access-list`
- [x] `client`
- [x] `compute-address`
- [x] `etherscan-source`
- [x] `help`
- [x] `index`
- [x] `interface`
- [x] `pretty-calldata`
- [x] `publish`
- [x] `receipt`
- [x] `sig`
I don't really want to open a pull request for each of these independently
I'm opening this draft pr now for anyone who wants to double check this all as I update.


onbjerg:
Ah, it seems we're doing double work. #185 completely restructures the cast command reference. Do you have an ETA on this? If it's not too long (few days), then we can merge this and I can split it out. In the meantime I'll work on Forge docs

zk-tarts:
I should be able to have these done *very* soon, probably today or tomorrow.

zk-tarts:
The commands I listed are all done.  
I still think its probably a good idea to go over every single `cast` subcommand to makes sure all the options are documented. 

If you'd like I can split them up (preserving the old formatting)
I'm not sure if you want to keep this old formatting style though seeing as the cheatcode reference was changed to be more like the `cargo` reference.

onbjerg:
@zk-tarts No, it's fine - I'll split them up and reuse whatever is in the file now and add what is missing (options etc) :) Let me know when you think this is good to merge

zk-tarts:
I just double checked everything, I'd say it's good to go now.

onbjerg:
Thank you for this :smile:

PR #186, opened 2022-04-08
add --private-key to deploy command in tutorial

djh58:

PR #187, opened 2022-04-08
Add private key arg to forge create command

0x1723:
This PR adds the `--private-key` arg to the `forge create` command in the example.

Without the private key arg, the command fails, as expected, since it needs the private key in order to deploy the contract. Here is the error shown when one is not provided:

```
Error:
   0: error accessing local wallet, did you set a private key, mnemonic or keystore? Run `cast send --help` or `forge create --help` and use the corresponding CLI flag to set your key via --private-key, --mnemonic-path, --interactive, --trezor or --ledger. Alternatively, if you're using a local node with unlocked accounts, set the `ETH_FROM` environment variable to the address of the account you want to use
```

My editor removed an empty space, I kept it and included it as part of the same commit, but I'm happy to either completely remove it and not change it, or remove it as part of another commit.


onbjerg:
Sorry, this was already in #186 which has been merged. Thank you for contributing :)

0x1723:
Oh I missed it @onbjerg 👍🏼

Issue #212, opened 2022-04-18
Document slither & mythril integrations


wminshew:
pulled from [this recent repo of mine](https://github.com/0xSplits/splits-vesting),

slither.config.json
```
{
  "filter_paths": "lib",
  "solc_remaps": [
    "ds-test/=lib/ds-test/src/",
    "forge-std/=lib/forge-std/src/",
  ]
}
```

mythril.config.json
```
{
  "remappings": [
    "ds-test/=lib/ds-test/src/",
    "forge-std/=lib/forge-std/src/",
  ],
  "optimizer": {
    "enabled": true,
    "runs": 200
  }
}
```

I struggled a bit to find/figure out the right config format (the original errors provided iirc were not particularly helpful in identifying it was a remapping issue, just a guess from my limited compiler knowledge). Have seen at least one other person recently on tg support group ask about it. Maybe good for an appendix page?

onbjerg:
I think it makes sense to include it in the configuration section, where we also have stuff like VSCode and GitHub actions :smile:

wminshew:
> I think it makes sense to include it in the configuration section, where we also have stuff like VSCode and GitHub actions 😄

totally :) opened a PR lmkwyt

Elyx0:
A bit confused about this since I followed the foundry docs and added

- Mythril version v0.23.4

```json
{
    "remappings": [
      "openzeppelin-contracts/=lib/openzeppelin-contracts/",
      "ds-test/=lib/ds-test/src/",
      "forge-std/=lib/forge-std/src/"
    ],
    "optimizer": {
      "enabled": true,
      "runs": 200
    }
  }
  ```
  in a `mythril.config.json` at the root of my repo. 

And still get:

```
mythril.interfaces.cli [ERROR]: Solc experienced a fatal error.

ParserError: Source "openzeppelin-contracts/contracts/utils/Address.sol" not found: File not found. Searched the following locations: "".
 --> contracts/FastLaneAuction.sol:4:1:
  |
4 | import "openzeppelin-contracts/contracts/utils/Address.sol";
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
  
  I did not find any references of  `mythril.config.json` in https://mythril-classic.readthedocs.io/en/master/search.html?q=mythril.config.json&check_keywords=yes&area=default nor a reference to `remappings`
  
[...]

PR #214, opened 2022-04-19
add static analyzers to config w slither & mythril

wminshew:
re https://github.com/foundry-rs/book/issues/212


onbjerg:
thanks! @all-contributors add @wminshew for docs

PR #216, opened 2022-04-19
fix: nft tutorial

dbelv:
…ssue


onbjerg:
thanks! @all-contributors add @dbelv for maintenance

PR #230, opened 2022-04-23
docs: fix `cast call` example command

Galmoli:
`cast call` example didn't have function name as string

### DIFF
--- a/scripts/generate-output-cast
+++ b/scripts/generate-output-cast
@@ -7,7 +7,7 @@ need_cmd cast
 echo "Generating output (cast)..."
 mkdir -p $OUTPUT_DIR/cast
 (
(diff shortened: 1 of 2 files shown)


onbjerg (changes requested):
You need to update it in this file: https://github.com/foundry-rs/book/blob/master/scripts/generate-output-cast

The file you edited is auto-generated

Galmoli:
I ran `generate-output-cast` and it generated the output correctly, so https://github.com/foundry-rs/book/blob/master/src/output/cast/cast-call has quotation marks.

gakonst (approved):

onbjerg:
@all-contributors add @Galmoli for maintenance

PR #231, opened 2022-04-24
add missing command

p0n1:
change directory into the cloned repo


onbjerg:
Thanks! @all-contributors add @p0n1 for maintenance

PR #239, opened 2022-04-25
feat(docs): import cheatcode tutorial

sambacha:
source: https://github.com/foundry-rs/foundry/commit/ddd6be4d8e25423a43d7a2c2af553e12a6e4d682


gakonst:
let's update this to use `import forge-std/Test.sol` and make the example work with `contract FooTest is Test`? And we can link to the cheatcodes reference instead of bringing in the entire `interface Vm`

gakonst:
ACK - closing and @sambacha if you want to add something, please extend q

sambacha:
image

PR #241, opened 2022-04-25
docs: add info about path to solc binary

mattsse:
add info about local solc installation

### DIFF
--- a/src/faq.md
+++ b/src/faq.md
@@ -41,7 +41,7 @@ Forge will try to auto-detect what Solidity compiler works for your project.
 To use a specific Solidity compiler, you can set [`solc`][config-solc] in your [config file][config],
 or pass `--use solc:<version>` to a Forge command that supports it (e.g. [`forge build`][forge-build]
 or [`forge test`][forge-test]). 
-Paths to a solc binary are also accepted. To use a specific local solc binary, set set path as `solc = "<path to solc>"` in your config file or pass `--use "<path to solc>"`. 
+Paths to a solc binary are also accepted. To use a specific local solc binary, you can set `solc = "<path to solc>"` in your config file, or pass `--use "<path to solc>"`. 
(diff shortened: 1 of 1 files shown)


gakonst (approved):
Should we also add a note about SOLC_PATH and the logging functionality?

mattsse:
added info about the `FOUNDRY_SOLC` env var.
should we info how to dump the std json in/out here are somewhere else @onbjerg ?

onbjerg:
@mattsse I think it would make sense to document that in the `forge build` command reference. Also, I don't think we have a `solc` heading in the config reference, would you mind adding that?

I also think the docs in `forge build` are somewhat lacking on the different ways to set solc, I think only `--use <version>` is documented atm

src/faq.md
```diff
@@ -38,9 +38,11 @@ If you want to run only a few tests, you can use `--match-test` to filter test f
 
 Forge will try to auto-detect what Solidity compiler works for your project.
 
-To use a specific Solidity compiler, you can set [`solc_version`][config-solc] in your [config file][config],
+To use a specific Solidity compiler, you can set [`solc`][config-solc] in your [config file][config],
 or pass `--use solc:<version>` to a Forge command that supports it (e.g. [`forge build`][forge-build]
-or [`forge test`][forge-test]).
+or [`forge test`][forge-test]). 
+Paths to a solc binary are also accepted. To use a specific local solc binary, set set path as `solc = "<path to solc>"` in your config file or pass `--use "<path to solc>"`.
```
onbjerg, line 44:
```suggestion
Paths to a solc binary are also accepted. To use a specific local solc binary, you can set `solc = "<path to solc>"` in your config file, or pass `--use "<path to solc>"`. 
```

gakonst (approved):
kk lets merge this small change and follow up w more `forge build`  docs

PR #225, opened 2022-04-22
Differential Testing and FFI

dmfxyz:
Added a write up on differential and standardized testing. Has examples of using `FFI` against non-evm executables, addressing #182.

Needs some editorial work, but content is about done so I am opening for review. It would be cool to include Frankie's GDA tests as an example as well, but avoiding doing that without permission/collaboration.

PR #244, opened 2022-04-26
Differential Testing: add link to Strings2 library for converting bytes to hex string

dmfxyz:
In case there are more questions like this: https://github.com/foundry-rs/book/pull/225#issuecomment-1110069116.

Also not sure if this would be useful in something like `forge-std` to help people with ffi testing? Happy to clean up the library and open a pr if we think it's useful.


gakonst (approved):
I could see string/bytes utils being useful in Forge std.

let's open an issue to discuss there, and only go for a PR when we agree what we want to add exactly?

PR #259, opened 2022-05-01
Fixed links and added "Forge Standard Library Reference" to references

Leutwiler:
1) The ffi link was broken. I redirected it to the place it should go for, which is https://book.getfoundry.sh/reference/config.html#ffi

2) Added the Forge Standard Library Reference name and link to the References part


onbjerg:
Thanks!! @all-contributors add @Leutwiler for docs, maintenance

onbjerg (approved):

PR #275, opened 2022-05-04
feat(installation.md): feat update to latest command

wiasliaw:
1. According to https://github.com/foundry-rs/foundry/pull/1037, current installation command will get error below.

```
error: multiple packages with binaries found: anvil, foundry-cli. When installing a git repository, cargo will always search the entire repo for any Cargo.toml. Please specify which to install.
```

2. provide update command to get the latest version of Foundry.

### DIFF
--- a/src/getting-started/installation.md
+++ b/src/getting-started/installation.md
@@ -47,8 +47,10 @@ Download and run `rustup-init` from [rustup.rs](https://rustup.rs). It will star
 After this, run the following to build Foundry from source:
 
 ```sh
-cargo install --git https://github.com/foundry-rs/foundry --bins --locked
+cargo install --git https://github.com/foundry-rs/foundry foundry-cli --bins --locked
 ```
+
+To update from source, run the same command again.
 
 ### Using with Docker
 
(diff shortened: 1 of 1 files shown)


onbjerg (changes requested):
Any way we can get it to install Anvil as well?

gakonst:
@mattsse should we pull in the Anvil binary into `foundry-cli` so that all CLI bins are in the same package and avoid this issue?

mattsse:
> @mattsse should we pull in the Anvil binary into `foundry-cli` so that all CLI bins are in the same package and avoid this issue?

if we do this then testing in `cli` will take get even longer.

I have no better solution yet, need to try something first.
the current setup of the `foundry-cli` is also not ideal, because it's essentially two crates in itself.

src/getting-started/installation.md
```diff
@@ -47,9 +47,11 @@ Download and run `rustup-init` from [rustup.rs](https://rustup.rs). It will star
 After this, run the following to build Foundry from source:
 
 ```sh
-cargo install --git https://github.com/foundry-rs/foundry --bins --locked
+cargo install --git https://github.com/foundry-rs/foundry foundry-cli --bins --locked
```
onbjerg, line 50:
```suggestion
cargo install --git https://github.com/foundry-rs/foundry foundry-cli anvil --bins --locked
```

onbjerg (commented):
To also install Anvil

PR #278, opened 2022-05-04
docs: add 2nd form of expectEmit to cheatcodes reference

wardbradt:
added comment matches line here https://github.com/foundry-rs/foundry/blob/b7c0774f3599f800dcf61c7b329761cce611eb57/testdata/cheats/Cheats.sol#L46


onbjerg:
@all-contributors add @wardbradt for maintenance

Issue #301, opened 2022-05-14
Document SMT options


onbjerg:
Requires changes in the config reference

See https://github.com/foundry-rs/foundry/pull/1602

shawnharmsen:
will take a stab at this one

PR #303, opened 2022-05-14
docs: document smt options

shawnharmsen:
issue: https://github.com/foundry-rs/book/issues/301

also added the additional optimizer settings because I didn't see it

https://github.com/foundry-rs/foundry/tree/master/config#additional-optimizer-settings

and

https://github.com/foundry-rs/foundry/tree/master/config#additional-model-checker-settings

### DIFF
--- a/src/reference/config.md
+++ b/src/reference/config.md
@@ -36,6 +36,8 @@ remappings = []
 libraries = []
 # Whether to cache builds or not
 cache = true
+# The cache directory if enabled
+cache_path = 'cache'
 # Whether to ignore the cache
 force = false
 # The EVM version by hardfork name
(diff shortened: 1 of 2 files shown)


onbjerg (changes requested):
The sections you wrote here are probably better for the `forge build` reference - whatever is in the config reference should follow the format that the rest of the document uses

shawnharmsen:
gotcha does this look like a better place to put it

0xvv (approved):
All good :+1:

onbjerg:
Yes, anything that is not in the config reference but is in Foundry is a candidate to be added; feel free to open up an issue or a PR with them :blush:

Issue #313, opened 2022-05-17
Document missing foundry.toml options in reference/config.md


shawnharmsen:
Filling in missing items from

config/src/lib.rs:
https://github.com/foundry-rs/foundry/blob/35f60f1e80348ada20e8b54e7a61bea0fe61c076/config/src/lib.rs#L82

```
via_ir 
revert_strings
optimizer_details
eth_rpc_url
etherscan_api_key
test_pattern
test_pattern_inverse
contract_pattern
contract_pattern_inverse
path_pattern
path_pattern_inverse
block_gas_limit
memory_limit
names
sizes

```

To the book:
https://book.getfoundry.sh/reference/config.html#configuration-format

PR #314, opened 2022-05-17
docs: add missing config reference

shawnharmsen:
fixes: https://github.com/foundry-rs/book/issues/313

1. Would like some feedback on the configuration keys

2. I left the following commented out in the `toml` like `#solc_version`
- wasn't sure how to represent `None` in toml

```toml
# Switch optimizer components on or off in detail
#optimizer_details = None
# Url of the rpc server that should be used for any rpc calls
#eth_rpc_url = None
# Etherscan API key
#etherscan_api_key = None
# Only run test functions matching the specified regex pattern
#test_pattern= None
# Only run test functions that do not match the specified regex pattern
#test_pattern_inverse = None
# Only run tests in contracts matching the specified regex pattern
#contract_pattern = None
# Only run tests in contracts that do not match the specified regex pattern
#contract_pattern_inverse = None
# Only run tests in source files matching the specified glob pattern
#path_pattern = None
# Only run tests in source files that do not match the specified glob pattern
#path_pattern_inverse = None
# The block gas limit
#block_gas_limit = None
# The memory limit of the EVM (32 MB by default)
```

### DIFF
--- a/src/reference/config.md
+++ b/src/reference/config.md
@@ -109,8 +109,34 @@ bytecode_hash = "ipfs"
 sparse_mode = false
 # Whether or not to use the Yul intermediate representation compilation pipeline
 via_ir = false
-# How to treat revert (and require) reason strings.
+# How to treat revert (and require) reason strings
 revert_strings = "default"
+# Switch optimizer components on or off in detail
+#optimizer_details = None
+# Url of the rpc server that should be used for any rpc calls
+#eth_rpc_url = None
+# Etherscan API key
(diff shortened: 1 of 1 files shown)


0xvv (commented):
Clean, i didn't find any mistake

0xvv:
There is no null value in TOML, i guess leaving it commented is the best option, the recommendation seems to be to include it only if needed.

onbjerg (approved):

Issue #316, opened 2022-05-18
Document `fail(string)` std assertion


ZeroEkkusu:
See https://github.com/foundry-rs/forge-std/pull/71

Issue #226, opened 2022-04-23
Document Std Assertions (Forge Std)


ZeroEkkusu:
See "STD-ASSERTIONS" in Forge Std's [`Test.sol`](https://github.com/foundry-rs/forge-std/blob/master/src/Test.sol).

PR #318, opened 2022-05-18
docs: fail string std assertion

shawnharmsen:
closes: https://github.com/foundry-rs/book/issues/316
closes: https://github.com/foundry-rs/book/issues/226

Will add the remaining Std assertions and set to ready for review


src/reference/forge-std/assertApproxEqRel.md
```diff
@@ -0,0 +1,40 @@
...
+
+Asserts `a` is approximately equal to `b` with delta in percentage (An 18 decimal fixed point number, where 1e18 == 100%).
+
+### Examples
+
+```solidity
+uint256 a = 100;
+uint256 b = 200;
+function testFail () external {
+    assertApproxEqRel(a, b, 400000000000000000);
```
ZeroEkkusu, line 31:
```suggestion
    assertApproxEqRel(a, b, 0.4e18);
```

ZeroEkkusu, line 31: This is cleaner - gives a better example of it should be used.

```suggestion
    assertApproxEqRel(a, b, 0.4e18);
```

src/reference/forge-std/fail.md
```diff
@@ -0,0 +1,23 @@
...
+### Description
+
+Provides a clean and readable way to fail tests if a certain branch or execution point is hit.
+
+### Examples
+
+```solidity
+function test() external {
+    fail("failed");
+}
```
ZeroEkkusu, line 18: Better example.

```suggestion
function test() external {
    for(int id; id < supply; ++i){
        if(collectible.ownerOf(id) == address(this)) return;
    }
    fail("Does not own any collectible.");
}
```

ZeroEkkusu, line 18: Better example.

```suggestion
function testOwned() external {
    for(int id; id < supply; ++i){
        if(collectible.ownerOf(id) == address(this)) return;
    }
    fail("Does not own a collectible.");
}
```

ZeroEkkusu (commented):
Could you also add it here as well:

- [Forge Standard Library Overview](https://book.getfoundry.sh/forge/forge-std.html#std-math)
- [Forge Standard Library Reference](https://book.getfoundry.sh/reference/forge-std/index.html#forge-stds-test)

src/reference/forge-std/assertEq.md
```diff
@@ -0,0 +1,24 @@
...
+```
+
+```solidity
+function assertEq(bytes memory a, bytes memory b, string memory err) internal;
+```
+
+
+### Description
+
+Assert `a` is equal to `b`.
```
0xvv, line 24: Asserts *

0xvv (commented):
nitpicking

shawnharmsen:
I think I got everything

ZeroEkkusu (approved):

onbjerg (approved):

Issue #332, opened 2022-05-29
Document Solidity Scripting


ZeroEkkusu:
See https://github.com/foundry-rs/foundry/pull/1208

Perelyn-sama:
I would like to mess around with this👉👈

ZeroEkkusu:
@Perelyn-sama Perhaps you can write a section about Solidity Scripting, after `Deploying and Verifying`. I think it deserves a dedicated section. Wdyt, @onbjerg?

PR #335, opened 2022-05-30
docs: add forge script

shawnharmsen:
closes: https://github.com/foundry-rs/book/issues/332

1. Should `forge script` be under `References/forge Commands/Deploy Commands` or `Utility Commands` like `forge run` ?
2. `forge run` should be changed to `forge debug` I think, maybe address in separate issue

I just noticed that `forge script` is using `multi_wallets.rs`
https://github.com/foundry-rs/foundry/blob/b192af49c6e1f3874bc6775cfd969040fddce740/cli/src/opts/multi_wallet.rs#L86

for its wallet options rather than `wallet.rs`
https://github.com/foundry-rs/foundry/blob/b192af49c6e1f3874bc6775cfd969040fddce740/cli/src/opts/wallet.rs#L52

~~I will update docs shortly~~

## changes
- added `src/reference/common/multi-wallet-options.md` with plural and different short flag for `--froms`


src/reference/forge/forge-script.md
```diff
@@ -0,0 +1,67 @@
...
+Run a smart contract as a script, building transactions that can be sent onchain.
+Adds the ability for forge to use a solidity file to create a contract on mainnet and arbitrary deploy scripts by adding a cheatcode.
+
+### OPTIONS
+
+`--broadcast`  
+&nbsp;&nbsp;&nbsp;&nbsp;Broadcasts the transactions.
+
+`--debug`  
+&nbsp;&nbsp;&nbsp;&nbsp;Takes precedence over broadcast.
```
onbjerg, line 22: What does it do? If this is the description of the flag in Forge we should fix that upstream as well as this is pretty non-descriptive

src/reference/forge/forge-script.md
```diff
@@ -0,0 +1,67 @@
...
+
+`--sig` *signature*  
+&nbsp;&nbsp;&nbsp;&nbsp;The signature of the function you want to call in the contract, or raw calldata.  
+&nbsp;&nbsp;&nbsp;&nbsp;Default: `run()`  
+
+`--slow`  
+&nbsp;&nbsp;&nbsp;&nbsp;Makes sure a transaction is sent, only after its previous one has been confirmed and succeeded.
+
+`--target-contract` *contract_name*  
+&nbsp;&nbsp;&nbsp;&nbsp;The name of the contract you want to run
```
onbjerg, line 38:
```suggestion
&nbsp;&nbsp;&nbsp;&nbsp;The name of the contract you want to run.
```

onbjerg (changes requested):
Minor nits, however I think we should include the `forge debug` change in this PR as well. I also think `forge script` should probably be under "Utility Commands" for now, since it sort of replaces `forge run` (which is now just a way to run the debugger)

src/reference/forge/forge-script.md
shawnharmsen: It opens it in the debugger.

same results for both commands below

```bash
forge script ./test/Broadcast.t.sol --tc BroadcastTest --sig "deploy()" -vvv --fork-url $GOERLI_RPC_URL --debug
forge debug ./test/Broadcast.t.sol --tc BroadcastTest --sig "deploy()"
```

shawnharmsen:
Think I got it all

1. Change `forge run` to `forge debug`
2. All `forge script` docs
3. `forge script --debug` help text in foundry pr: https://github.com/foundry-rs/foundry/pull/1782

onbjerg (approved):

Issue #329, opened 2022-05-27
Document `cast upload-signature`


onbjerg:
This is a utility function that uploads a selector to https://sig.eth.samczsun.com

See https://github.com/foundry-rs/foundry/pull/1716

PR #353, opened 2022-06-02
docs: add cast upload-signature

shawnharmsen:
closes: https://github.com/foundry-rs/book/issues/329

Added it under cast - abi commands: https://book.getfoundry.sh/reference/cast/abi-commands.html


0xvv (commented):
looks good to me

PR #393, opened 2022-06-22
nit: fix cast block command in docker tutorial

dmfxyz:
Missing the `--block` tag. I know we are considering changing this as a requirement in the command to mimic seth, but for now it should be updated.


onbjerg:
It was already changed, so we'd have to revert this tonight :smile: See #394

Issue #391, opened 2022-06-22
Dcoument `--base-fee` alias


onbjerg:
`--base-fee-per-gas` in Anvil and Forge now has an alias: `--base-fee`.

See https://github.com/foundry-rs/foundry/pull/2058

PR #399, opened 2022-06-23
add base-fee alias

0xvv:
closes https://github.com/foundry-rs/book/issues/391

### DIFF
--- a/src/reference/anvil/README.md
+++ b/src/reference/anvil/README.md
@@ -293,7 +293,7 @@ Returns the details of all transactions currently pending for inclusion in the n
 
 
 #### Executor Environment Config
-`--block-base-fee-per-gas <FEE>`  
+`--base-fee, --block-base-fee-per-gas <FEE>`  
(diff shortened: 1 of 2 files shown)


src/reference/anvil/README.md
```diff
@@ -293,7 +293,7 @@ Returns the details of all transactions currently pending for inclusion in the n
 
 
 #### Executor Environment Config
-`--block-base-fee-per-gas <FEE>`  
+`--base-fee, --block-base-fee-per-gas <FEE>`
```
onbjerg, line 296:
```suggestion
`--base-fee`  
`--block-base-fee-per-gas <FEE>`  
```

Same for the other change

onbjerg (changes requested):

src/reference/anvil/README.md
0xvv: It does look better, i was trying to stay consistent with what was above, but those were one letter aliases

onbjerg (approved):

PR #401, opened 2022-06-25
Update forge-completions.md

sameul2012:
Add forge completions for docker.


onbjerg:
Unsure if this fits in the book - these steps largely depend on what Docker base image you are using. If we want this for the one that we release via CI I would push either for:

- Adding these completions by default in the Docker image
- Moving these docs to the Docker image tutorial

gakonst:
closing per @onbjerg's comment

PR #406, opened 2022-06-28
docs: document struct coded and add examples

mattsse:
document struct ABI codec and add examples to cast (sig, send)

ref https://github.com/foundry-rs/foundry/pull/2151


onbjerg:
I think the struct encoding file needs to be in summary.md to be rendered

mattsse:
was the line indent lol

gakonst:
Can we also make it work so that people can copy/paste struct definitions in line? `cast call 0x1234 `struct Foo { address x; int256 y; }` or do we think it doesn't make sense?

src/misc/README.md
```diff
@@ -0,0 +1,3 @@
+## Misc
+
+- [struct encoding](./struct-encoding.md)
```
onbjerg, line 3:
```suggestion
- [Struct encoding](./struct-encoding.md)
```

src/misc/struct-encoding.md
```diff
@@ -0,0 +1,64 @@
+#### Struct Encoding
```
onbjerg, line 1:
```suggestion
## Struct Encoding
```

src/misc/struct-encoding.md
```diff
@@ -0,0 +1,64 @@
...
+
+Structs are custom defined types that can group several variables:
```
onbjerg, line 3:
```suggestion
Structs are user defined types that can group several variables:
```

src/misc/struct-encoding.md
```diff
@@ -0,0 +1,64 @@
...
+```solidity
+struct MyStruct {
+    address addr;
+    uint256 amount;
+}
+```
+
+Only the new [ABI coder v2](https://docs.soliditylang.org/en/latest/layout-of-source-files.html#abi-coder-pragma) can encode and decode arbitrarily nested arrays and structs. Since Solidity 0.8.0 it is activated by default, prior to that it needs to be activated via `pragma experimental ABIEncoderV2`.
+
+Structs map to the ABI type "tuple", See also [Mapping Solidity to ABI types](https://docs.soliditylang.org/en/latest/abi-spec.html#mapping-solidity-to-abi-types)
```
onbjerg, line 14:
```suggestion
Solidity structs map to the ABI type "tuple". For more information on how Solidity types map to ABI types see [Mapping Solidity to ABI types](https://docs.soliditylang.org/en/latest/abi-spec.html#mapping-solidity-to-abi-types) in the Solidity documentation.
```

src/misc/struct-encoding.md
```diff
@@ -0,0 +1,64 @@
...
+
+Structs are therefor encoded and decodes as tuples. So `MyStruct` is a `(address,uint256)` tuple when it comes to the ABI spec.
```
onbjerg, line 16:
```suggestion
Structs are therefore encoded and decodes as tuples. So the struct we defined above, `MyStruct`, maps to the tuple `(address,uint256)` in terms of the ABI.
```

src/misc/struct-encoding.md
```diff
@@ -0,0 +1,64 @@
...
+
+For example the ABI of
```
onbjerg, line 18:
```suggestion
Let's see how this works in a contract:
```

src/misc/struct-encoding.md
```diff
@@ -0,0 +1,64 @@
...
+contract Test {
+    struct MyStruct {
+        address addr;
+        uint256 amount;
+    }
+    function f(MyStruct memory t) public pure {}
+}
+```
+
+results in the following ABI json object:
```
onbjerg, line 33:
```suggestion
The ABI of the `f` function in this contract is:
```

src/misc/struct-encoding.md
```diff
@@ -0,0 +1,64 @@
...
+	],
+	"name": "f",
+	"outputs": [],
+	"stateMutability": "pure",
+	"type": "function"
+}
+```
+
+which reads:
+The function `f` takes 1 input of type `tuple` with two components of type `address` and `uint256`.
```
onbjerg, line 64:
```suggestion
which reads: The function `f` takes 1 input of type `tuple` with two components of type `address` and `uint256`.
```

src/reference/cast/cast-send.md
```diff
@@ -52,6 +52,23 @@ The destination (*to*) can be an ENS name or an address.
...
+        }
+        function myfunction(MyStruct memory t) public pure {}
+    }
+    ```
+
+Structs are encoded as tuples, See [struct encoding](./reference/common/struct-encoding.md)
+
+```sh
+cast send --ledger 0x... "myfunction((address,uint256))" 0x... 1
+```
```
onbjerg, line 71:
````suggestion
    Structs are encoded as tuples (see [struct encoding](./reference/common/struct-encoding.md))

    ```sh
    cast send --ledger 0x... "myfunction((address,uint256))" 0x... 1
    ```
````

Also, would this not be e.g. `cast send "myfunction((address,uint256))" 0x... "(0x..., 1)"`?

File: src/reference/cast/cast-sig.md [hunk 10]
```diff
@@ -25,6 +25,23 @@ The signature (*sig*) is a fragment in the form `<function name>(<types...>)`.
...
+        }
+        function myfunction(MyStruct memory t) public pure {}
+    }
+    ```
+
+Structs are encoded as tuples, See [struct encoding](./reference/common/struct-encoding.md)
+
+```sh
+cast sig "myfunction((address,uint256))"
+```
```
onbjerg, line 44:
````suggestion
    Structs are encoded as tuples (see [struct encoding](./reference/common/struct-encoding.md)).

    ```sh
    cast sig "myfunction((address,uint256))"
    ```
````

**Review by onbjerg — CHANGES_REQUESTED** [review]
Got some structural nits/phrasing but also have one comment for one of the examples. Will just batch commit the nits except for the one I'm unsure about

File: src/reference/cast/cast-send.md (hunk 9 above)
Comment (mattsse) [reply src/reference/cast/cast-send.md]:
```console

USAGE:
    cast send [OPTIONS] <TO> [ARGS]

ARGS:
    <TO>
            The destination of the transaction.

    <SIG>
            The signature of the function to call.

    <ARGS>...
            The arguments of the function to call.

```

so `cast send <to> <sig> <args>` is correct here

onbjerg: Ye but args in this case would be "(0x..., 1)". The current example is:

```
cast send --ledger 0x... "myfunction((address,uint256))" 0x... 1
```

In other words:

Send using Ledger addr 0x..., call "myfunction" at address 0x... with argument 1?

mattsse: guess that makes sense, I only copy pasted the example tbh

onbjerg (approved):

onbjerg:
Can merge when the base PR is merged

PR #408, opened 2022-06-29
chore: adds Docker setup instructions

omarish:
This PR adds a Dockerfile to start `mdbook`. 

# Why?

I saw that the the main book URL [was down](https://twitter.com/gakonst/status/1542175547731480583) and I wanted to spin up a mirror.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM rust
+RUN cargo install mdbook
+EXPOSE 3000
+COPY . /book
+WORKDIR /book
+ENTRYPOINT ["mdbook", "serve", "-n", "0.0.0.0"]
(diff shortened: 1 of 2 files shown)


mattsse (commented):
there's also https://github.com/peaceiris/docker-mdbook

but I think this will do it.

should we add a workflow to publish this to ghcr @onbjerg?

onbjerg:
I don't really think this makes a lot of sense, I would just defer to the docker-mdbook image linked above

mattsse:
right, how about adding instructions for serving with `docker-mdbook` instead @omarish ?

onbjerg (approved):
@all-contributors add @omarish for docs

PR #422, opened 2022-07-06
Add a missing step for installing the forge-std library

ebaizel:
I needed to install the standard library in order to `forge build` the demo repo.


onbjerg:
Are you on the latest Foundry? The latest Foundry should install Forge Std for you when you run `forge init`.

ebaizel:
Turns out I have a local git defender check that interferes with forge-std being built. Once I set up the git defender, I was able to successfully install with `forge init hello_foundry --force --no-commit`

PR #441, opened 2022-07-11
add computeCreateAddress reference to book

devanonon:
After adding this function to the std library I figured I should update the book.

https://github.com/foundry-rs/forge-std/pull/111


mattsse (approved):
awesome, ty!

PR #442, opened 2022-07-11
docs: document forking related cheatcodes

mattsse:
documents new cheatcodes for https://github.com/foundry-rs/foundry/pull/1715

### DIFF
--- a/src/cheatcodes/snapshots.md
+++ b/src/cheatcodes/snapshots.md
@@ -1,0 +1,57 @@
+## snapshot cheatcodes
+
+### Signature
+
+```solidity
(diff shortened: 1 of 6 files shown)


src/cheatcodes/snapshots.md
```diff
@@ -0,0 +1,57 @@
...
+### Examples
+
+
+```solidity
+struct Storage {
+    uint slot0;
+    uint slot1;
+}
+
+contract SnapshotTest is DSTest {
```
gakonst, line 31: Minor thing - let's start replacing DSTest everywhere in the book with `forge-std/Test`?

PR #480, opened 2022-07-26
fix: links

pfives:
Link to remapping-dependencies was broken. Link to config was pointing to the html page.


mattsse (approved):
nice catch!

PR #485, opened 2022-07-27
Add sample travis config

sourabhmarathe:
This took a bit of experimentation to get working on my end so I figured it would be useful to provide this information for how to set this up for Travis as well.

### DIFF
--- a/src/config/continous-integration.md
+++ b/src/config/continous-integration.md
@@ -32,7 +32,7 @@ jobs:
 To test your project using Travis CI, here is a sample workflow:
 
 ```yml
-lanaguage: rust
+language: rust
 cache:
   cargo: true
   directories:
--- a/src/config/continous-integration.md
+++ b/src/config/continous-integration.md
@@ -26,3 +26,23 @@ jobs:
       - name: Run tests
         run: forge test -vvv
 ```
+
+### Travis CI
+
+To test your project using Travis CI, here is a sample workflow:
+
+```yml
+lanaguage: rust
+cache:
+  cargo: true
+  directories:
+    - $HOME/.foundry
+
+install:
+  - curl -L https://foundry.paradigm.xyz | bash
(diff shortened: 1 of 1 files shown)


src/config/continous-integration.md
```diff
@@ -26,3 +26,23 @@ jobs:
...
+lanaguage: rust
+cache:
+  cargo: true
+  directories:
+    - $HOME/.foundry
+
+install:
+  - curl -L https://foundry.paradigm.xyz | bash
+  - export PATH=$PATH:$HOME/.foundry/bin
+  - foundryup -b master
```
mattsse, line 44: this will install from master instead of downloading and running static bins.

does this work out of the books with `foundryup`? 


@onbjerg thoughts an adding travis CI docs? would mean that we need to ensure this works on travis, haven't used it in years.

sourabhmarathe: It does work with `foundryup`, but I had to install from master in order to avoid a compilation error that `foundryup` can run into [here](https://github.com/foundry-rs/foundry#out-of-date-glibc-error-when-running-forge-from-default-foundryup-install).

I am open to doing the later, but I am not sure how that would work. Instead of `master`, would a release branch name be listed instead?

mattsse: sgtm, better to have it than not have a travis example

mattsse (approved):

src/config/continous-integration.md
```diff
@@ -26,3 +26,23 @@ jobs:
       - name: Run tests
         run: forge test -vvv
 ```
+
+### Travis CI
+
+To test your project using Travis CI, here is a sample workflow:
+
+```yml
+lanaguage: rust
```
gakonst, line 35:
```suggestion
language: rust
```

gakonst (approved):

PR #487, opened 2022-07-29
fix(rpc.md): error config in rpc_endpoints

wiasliaw:
Correct the config according to [Config](https://book.getfoundry.sh/reference/config#rpc_endpoints).


mattsse (approved):
ty!

PR #524, opened 2022-08-13
mention that slot number should be hex

skaunov:
https://github.com/foundry-rs/foundry/issues/2738#issuecomment-1213055370

### DIFF
--- a/src/reference/cast/cast-storage.md
+++ b/src/reference/cast/cast-storage.md
@@ -10,7 +10,7 @@ cast-storage - Get the raw value of a contract's storage slot.
 
 ### DESCRIPTION
 
-Get the raw value of a contract's storage slot.
+Get the raw value of a contract's storage slot given as hex.
 
 The address (*address*) can be an ENS name or an address.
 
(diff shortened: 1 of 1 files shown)


mattsse (commented):
I'm not sure this is still accurate 

could you check the current nightly. If that's still the case then we should rather add support for hex+numeric for that command

ref https://github.com/foundry-rs/foundry/pull/2741

src/reference/cast/cast-storage.md
```diff
@@ -10,7 +10,7 @@ cast-storage - Get the raw value of a contract's storage slot.
 
 ### DESCRIPTION
 
-Get the raw value of a contract's storage slot.
+Get the raw value of a contract's storage slot. (Slot numbers after 18446744073709551615 (Rust u64) should be given as hex.)
```
mattsse, line 13:
```suggestion
Get the raw value of a contract's storage slot. (Slot numbers greater than 18446744073709551615 (u64::MAX) should be given as hex.)
```

mattsse (changes requested):

skaunov:
Indeed, it wasn't. Just added info that hex should be passed for slots gt Rust u64 according to Onbjerg explanation.
> I'm not sure this is still accurate

mattsse (approved):

Issue #523, opened 2022-08-13
Document projectRoot cheatcode


mattsse:
ref https://github.com/foundry-rs/foundry/pull/2762

PR #525, opened 2022-08-13
Add projectRoot cheatcode page

jdgc:
Adds a page for the cheatcode added in https://github.com/foundry-rs/foundry/pull/2762

This PR closes https://github.com/foundry-rs/book/issues/523

Simple feature, so not much to say here.


mattsse (approved):
awesome, ty!

PR #534, opened 2022-08-15
Update fmt config options

rkrasiuk:
Update `FormatConfig` options per [source](https://github.com/foundry-rs/foundry/blob/44cd2008351b6b62ddb28d850f629f3073a576b3/config/src/fmt.rs)


onbjerg (approved):
thank you! :smile:

PR #535, opened 2022-08-17
Fix compilation issue in SimpleStorageContract

azanux:
The code does not compile, I made some modifications so that the beginners can make it compile without too much effort

By replacing "cheats" by "vm" at the right place in the code.
and "string public value;" by "uint256 public value;" in the smart contract SimpleStorageContract.


mattsse (approved):
good point, ty!

onbjerg:
@all-contributoirs add @azanux for maintenance

onbjerg:
@all-contributors add @azanux for maintenance

PR #538, opened 2022-08-18
Nit on foundry.toml profile

ControlCplusControlV:
Updated the foundry.toml section to now be in line with how this should be written in the lastest version of foundry


mattsse (approved):

onbjerg:
@all-contributors add @ControlCplusControlV for maintenance

PR #541, opened 2022-08-18
Fix Slither Static Analyzer Config Json usage syntax

m0ham3dx:
# Description 

Changes to -  `https://github.com/foundry-rs/book/blob/master/src/config/static-analyzers.md`

Addition of descriptive text and correct syntax for the usage of `slither` with a custom `file-config.json` and also includes the raw ouput, after a new repository has initialized with `forge init`


mattsse (approved):
ty!

PR #542, opened 2022-08-19
Fix mythril syntax and additions add slither github action link for ci/cd

m0ham3dx:
# Description 

This pull request has 3actions in two commits 

Commit 1 - mythril custom json usage syntax correction & addition of example output for mythril 
Commit 2 - Adding github action which are available for incorporating slither into CI/CD

### DIFF
--- a/src/config/static-analyzers.md
+++ b/src/config/static-analyzers.md
@@ -67,7 +67,48 @@ Note, you need switch `rustc` to nightly to install `mythril`:
 ```ignore
 rustup default nightly
 pip3 install mythril
-myth analyze src/Contract.sol --sol-json mythril.config.json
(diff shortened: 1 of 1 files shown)


mattsse (approved):
appreciate it!

onbjerg:
@all-contributors add @m0ham3dx for maintenance

m0ham3dx:
@onbjerg - 

The syntax to generate the illustrated output is missing a `-v` flag [HERE](https://github.com/foundry-rs/book/blob/master/src/config/static-analyzers.md?plain=1#L80) 

the code should be as per the initial commit 

```bash 
myth -v analyze src/Counter.sol --solc-json mythril.config.json
```

PR #543, opened 2022-08-19
fix: clarify default location for contracts

AbhiGulati:
Contracts in forge projects are in the `src` directory. Would be clearer to say that instead of the word "source".


onbjerg (approved):

onbjerg:
@all-contributors add @AbhiGulati for maintenance

PR #554, opened 2022-08-24
Remove --block-gas-limit from README.md

khelmy:
The `--block-gas-limit` flag doesn't seem to work as a CLI argument for anvil (at least on macOS). The `--gas-limit` flag works and is already included below. This PR just removes mention of `--block-gas-limit`.


onbjerg (approved):
@all-contributors add @khelmy for maintenance

onbjerg:
@all-contributors add @khelmy for maintenance

PR #569, opened 2022-08-28
Usage of sub commands with the snapshot command

HashHaran:
I tried out of curiosity if I can filter based on match path and match test when taking snapshot. It makes so much sense to have this feature. As expected it worked. Adding it in the documentation to help a future reader :)

### DIFF
--- a/src/forge/gas-snapshots.md
+++ b/src/forge/gas-snapshots.md
@@ -40,7 +40,7 @@ you can use the `--max <VALUE>` option.
 Keep in mind that the changes will be made in the snapshot file, and not in the snapshot being
 displayed on your screen.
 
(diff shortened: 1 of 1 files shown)


src/forge/gas-snapshots.md
```diff
@@ -40,6 +40,8 @@ you can use the `--max <VALUE>` option.
 Keep in mind that the changes will be made in the snapshot file, and not in the snapshot being
 displayed on your screen.
 
+You can also use it in combination with other subcommands, such as forge snapshot --match-path contracts/test/ERC721.t.sol, to generate a gas snapshot relevant to this test contract.
```
onbjerg, line 43:
```suggestion
You can also use it in combination with the filters for `forge test`, such as `forge snapshot --match-path contracts/test/ERC721.t.sol` to generate a gas snapshot relevant to this test contract.
```

onbjerg (approved):

onbjerg:
@all-contributors add @HashHaran for docs

PR #571, opened 2022-08-29
Fixed code snippet that doesn't compile, typos

robriks:
line 195 simple.value() needed a ; at end of line to compile
little typos here and there:
'therefor' -> 'therefore'
'contract does not exists' -> 'contract does not exist'
'contract only lifes in "mainnetFork"' -> 'contract only lives in "mainnetFork"'
persitent -> persistent


mattsse (approved):
ty!

Issue #588, opened 2022-09-05
Document forge verify-contract --constructor-args-path


0xYYY:
See https://github.com/foundry-rs/foundry/pull/3080

PR #589, opened 2022-09-05
docs: forge verify-contract --constructor-args

0xYYY:
Closes https://github.com/foundry-rs/book/issues/588.


mattsse:
smol test error, should be fixed by marking

```
    ForgeUSD FUSD 18 1000000000000000000000
```

as ```text I think

0xYYY:
thanks! it's fixed now

mattsse (approved):

PR #596, opened 2022-09-06
docs: add references and examples for rememberKey and deriveRememberKey

devanonon:
Re https://github.com/foundry-rs/foundry/pull/3047

Will follow-up with updates to the "Solidity Scripting" tutorial and security implications.


mattsse (approved):
awesome! ty

PR #602, opened 2022-09-07
fix(expect-revert): fix usage of reason string

wiasliaw:
Reason string will be encode as output, use `bytes(reason string)` will not pass the test.

### DIFF
--- a/src/cheatcodes/expect-revert.md
+++ b/src/cheatcodes/expect-revert.md
@@ -42,7 +42,7 @@ There are 3 signatures:
 >
 > ```solidity
 > function testLowLevelCallRevert() public {
->     vm.expectRevert(bytes("error message"));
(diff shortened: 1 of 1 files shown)


onbjerg (changes requested):
I think this is also not entirely correct - you don't need to use `abi.encode` or `bytes()`, at least we don't do that in our tests: https://github.com/foundry-rs/foundry/blob/master/testdata/cheats/ExpectRevert.t.sol (see eg https://github.com/foundry-rs/foundry/blob/53d286d44b4f30182b36c4330ee958d328432222/testdata/cheats/ExpectRevert.t.sol#L73-L77)

wiasliaw:
@onbjerg My fault, I found my typo was wrong.

Issue #615, opened 2022-09-09
document `[fuzz]max_test_rejects` in config


0xYYY:
See https://github.com/foundry-rs/foundry/pull/3138.

PR #616, opened 2022-09-10
doc: config fuzz.max_test_rejects

0xYYY:
Closes https://github.com/foundry-rs/book/issues/615.

Also took this chance to update `config.default.toml` to better match the output of `forge config`.


mattsse (approved):
awesome, tysm!

PR #617, opened 2022-09-10
docs: add cast --to-base

DaniPopes:
ref foundry-rs/foundry#3161
`--to-dec` and `--to-hex` were aliased to `--to-base`


mattsse (approved):
sweet! tysm

PR #642, opened 2022-09-21
docs: document allowCheatcodes

mattsse:
document `allowCheatcodes` cheatcode 

https://github.com/foundry-rs/foundry/pull/3042/files


src/cheatcodes/allow-cheatcodes.md
```diff
@@ -0,0 +1,19 @@
...
+
+```solidity
+function allowCheatcodes(address) external;
+```
+
+### Description
+
+In forking mode, explicitly grant the given address cheatcode access.
+
+By default, the test contract, and it's deployer are allowed to access cheatcodes. In addition to that, cheat code
```
mattsse, line 13:
```suggestion
By default, the test contract, and its deployer are allowed to access cheatcodes. In addition to that, cheat code
```

PR #645, opened 2022-09-22
remove redundant dash from ---extra-output-files

vvakame:
I met an error

```
error: Found argument '---extra-output-files' which wasn't expected, or isn't valid in this context

	Did you mean '--extra-output-files'?
```


mattsse (approved):

mattsse:
ty!

PR #656, opened 2022-09-28
Small docs options updates.

### LINKED ISSUE #650: Mnemonic options in documentation are no longer up to date.
The options in this file are wrong: https://github.com/foundry-rs/book/blob/98f5e0e7a24009976560c5c88ad4685ddb2ce4bc/src/reference/common/wallet-options-raw.md

The output of forge from my install today is the following:
```
WALLET OPTIONS - RAW:
    -i, --interactives <NUM>
            Open an interactive prompt to enter your private key. Takes a value for the number of keys to enter
            
            [default: 0]

        --mnemonic-derivation-paths <PATHS>
            The wallet derivation path. Works with both --mnemonic-path and hardware wallets.

        --mnemonic-indexes <INDEXES>
            Use the private key from the given mnemonic index. Used with --mnemonic-paths.
            
            [default: 0]

        --mnemonic-passphrases <PASSPHRASE>
            Use a BIP39 passphrases for the mnemonic.

        --mnemonics <PATHS>
            Use the mnemonic phrases or mnemonic files at the specified paths.

        --private-key <RAW_PRIVATE_KEY>
            Use the provided private key.

        --private-keys <RAW_PRIVATE_KEYS>
            Use the provided private keys.
```

laudiacay:
This PR updates some of the documentation for command line options to be in line with the current release of foundry.

Closes #650. 

Also has some extra little fixes (mostly for wallet-related options), but I absolutely did not get every discrepancy.

From what I saw, I think the best approach might be to have someone eventually implement a bash script that autogenerates docs from parsing the output of various `--help` calls, and plug that into CI... the docs have diverged very significantly from what's actually in the binary, and it'd be such a headache to have people keep manually updating them...

### DIFF
--- a/src/reference/common/display-options.md
+++ b/src/reference/common/display-options.md
@@ -2,4 +2,4 @@
 
 `-j`  
 `--json`  
-&nbsp;&nbsp;&nbsp;&nbsp;Output in JSON format.
+&nbsp;&nbsp;&nbsp;&nbsp; Print the deployment information as JSON.
--- a/src/reference/common/verifier-options.md
+++ b/src/reference/common/verifier-options.md
@@ -2,4 +2,5 @@
 &nbsp;&nbsp;&nbsp;&nbsp;The verification provider. Available options: `etherscan`, `sourcify` & `blockscout`. Default: `etherscan`.
 
 `--verifier-url` *url*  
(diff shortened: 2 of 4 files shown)


mattsse (approved):
thanks for adding these!

PR #666, opened 2022-10-07
Update solidity-scripting.md

rev3ks:
Instructions imply that the `startBroadcast` function takes a uint256 rather than an address

```
    // Has all subsequent calls (at this call depth only) create transactions that can later be signed and sent onchain
    function startBroadcast(address) external;
```

### DIFF
--- a/src/tutorials/solidity-scripting.md
+++ b/src/tutorials/solidity-scripting.md
@@ -158,7 +158,8 @@ import "../src/NFT.sol";
 contract MyScript is Script {
     function run() external {
         uint256 deployerPrivateKey = vm.envUint("PRIVATE_KEY");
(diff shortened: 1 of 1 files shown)


mattsse (commented):
hmm, there are functions for both:

https://github.com/foundry-rs/forge-std/blob/master/src/Vm.sol#L113-L116

but should probably be mentioned here

rev3ks:
> hmm, there are functions for both:
> 
> https://github.com/foundry-rs/forge-std/blob/master/src/Vm.sol#L113-L116
> 
> but should probably be mentioned here

Looks like my version is out of date so I don't have that function signature. 
```
   add41d3..2a2ce36  master                -> origin/master
```

PR #669, opened 2022-10-12
Update hardhat.md

zarifpour:
Added resolution to error: `cannot find name 'fs'.` when `getRemappings()` function is added to a file.
Step 3 of the documentation does not indicate that `fs` should be imported.


mattsse (approved):
lgtm

PR #674, opened 2022-10-21
Update wallet-options-raw.md

skaunov:
Please, review this one -- I didn't dig deep at all. Just updated in align with what freshly updated `cast` proposed when rejecting current variants of the commands.


mattsse (approved):
nice catch, yep these are the prefered variants.

tysm

Issue #629, opened 2022-09-19
Document verifier opts on `forge script`


onbjerg:
See https://github.com/foundry-rs/foundry/pull/3175

PR #676, opened 2022-10-21
verifier-options.md

shawnharmsen:
coming back from a hiatus so please bear with me as I get back up to speed!

fixes: https://github.com/foundry-rs/book/issues/629


mattsse (approved):
ty!

PR #686, opened 2022-10-27
doc: add best practices

mds1:
Leaving this as a draft to collect feedback on the content, cc @odyslam @lucas-manuel @gakonst

### DIFF
--- a/src/tutorials/best-practices.md
+++ b/src/tutorials/best-practices.md
@@ -1,0 +1,200 @@
+# Best Practices
+
+This guide documents the suggested best practices when developing with Foundry.
+In general, it's recommended to handle as much as possible with [`forge fmt`](../reference/config/formatter.md), and anything this doesn't handle is below.
+
+- [General Contract Guidance](#general-contract-guidance)
+- [Tests](#tests)
+  - [Test Harnesses](#test-harnesses)
+    - [Internal Functions](#internal-functions)
+    - [Private Functions](#private-functions)
+    - [Workaround Functions](#workaround-functions)
+  - [Best practices](#best-practices)
+  - [Taint Analysis](#taint-analysis)
+- [Scripts](#scripts)
+- [Comments](#comments)
+- [Resources](#resources)
+
+## General Contract Guidance
+
+1. Always use "absolute" paths for imports, i.e. paths relative to the repo root. This makes it easier to see where files are from and reduces churn when moving files around.
+
+   - Good: `import "script/Deploy.s.sol"`
+   - Bad: `import "../script/Deploy.s.sol"`
+
+1. Always use named import syntax, don't import full files. This restricts what is being imported to just the named items, not everything in the file. Importing full files can result in solc complaining about duplicate definitions and slither erroring, especially as repos grow and have more dependencies with overlapping names.
+
+   - Good: `import {MyContract} from "src/MyContract.sol"` to only import `MyContract`.
+   - Bad: `import "src/MyContract.sol"` imports everything in `MyContract.sol`. (Importing `forge-std/Test` or `Script` can be an exception here, so you get the console library, etc).
+
+1. Sort imports by `forge-std/` first, then dependencies, `test/`, `script/`, and finally `src/`. Within each, sort alphabetically by path (not by the explicit named items being imported). _(Note: This can be removed once [Foundry#3396](https://github.com/foundry-rs/foundry/issues/3396) is merged)._
+
+1. Similarly, sort named imports. _(Note: This can be removed once [Foundry#3396](https://github.com/foundry-rs/foundry/issues/3396) is merged)._
+
+   - Good: `import {bar, foo} from "src/MyContract.sol"`
+   - Bad: `import {foo, bar} from "src/MyContract.sol"`
+
+1. If copying a library from a dependency (instead of importing it), use the `ignore = []` option in the config file to avoid formatting that file. This makes diffing it against the original simpler for reviewers and auditors.
+
+1. Similarly, feel free to use the `disable-next-item`, `disable-next-line`, `disable-start`, `disable-end` code comment flags to ignore lines/sections of code that look better with manual formatting.
+
+Additional best practices from [samsczun](https://twitter.com/samczsun)'s [How Do You Even Write Secure Code Anyways](https://www.youtube.com/watch?v=Wm3t8Fuiy1E) talk:
+
+- Use descriptive variable names.
+- Limit the number of active variables.
+- No redundant logic.
+- Early exit as much as possible to reduce mental load when seeing the code.
+- Related code should be placed near each other.
+- Delete unused code.
+
+## Tests
+
+1. For testing `MyContract.sol`, the test file should be `MyContract.t.sol`. For testing `MyScript.s.sol`, the test file should be `MyScript.t.sol`.
+
+   - If the contract is big and you want to split it over multiple files, just group them logically like `MyContract.owner.t.sol`, `MyContract.deposits.t.sol`, etc.
+
+1. Never make assertions in the `setUp` function, instead use a dedicated test like `test_SetUpState()` if you need to ensure your `setUp` function does what you expected. More info on why in [Foundry#1291](https://github.com/foundry-rs/foundry/issues/1291)
+
+1. For unit tests, treat contracts as `describe` blocks:
+
+   - `contract Add` holds all unit tests for the `add` method.
+   - `contract Supply` holds all tests for the `supply` method.
+   - `contract Constructor` hold all tests for the constructor.
+   - One benefit of this approach is that smaller contracts should compile faster than large ones, so this approach of many small contracts should save time as test suites get large.
+
+1. Integration tests should live in the same `test` directory, with a clear naming convention. These may be in dedicated files, or they may live next to related unit tests in existing test files.
+
+1. Be consistent with test naming, as it's helpful for filtering tests (e.g. for gas reports you might want to filter out revert tests). When combining naming conventions, keep them alphabetical.
+
+   - `test_Description` for standard tests.
+   - `testFuzz_Description` for fuzz tests.
+   - `test_Revert[If|When]_Condition` for tests expecting a revert.
+   - `testFork_Description` for tests that fork from a network.
+   - `testForkFuzz_Revert[If|When]_Condition` for a fuzz test that forks and expects a revert.
+
+1. Be careful with with fuzz tests on a fork to avoid burning through RPC requests with non-deterministic fuzzing.
+
+1. When using assertions like `assertEq`, consider leveraging the last string param to make it easier to identify failures. These can be kept brief, or even just be numbers&mdash;they basically serve as a replacement for showing line numbers of the revert, e.g. `assertEq(x, y, "1")` or `assertEq(x, y, "sum1")`. _(Note: [Foundry#2328](https://github.com/foundry-rs/foundry/issues/2328) tracks integrating this natively)._
+
+1. Don't forget invariant tests! For the assertion string, use a verbose english description of the invariant: `assertEq(x + y, z, "Invariant violated: the sum of x and y must always equal z")`.
+
+1. When writing fork tests, do not use the `--fork-url` flag. Instead:
+
+   - Define `[rpc_endpoints]` in the config file and use the `createSelectFork` family of cheatcodes.
+   - Always pin to a block so tests are deterministic and RPC responses are cached.
+   - More info on this fork test approach [here](https://twitter.com/msolomon44/status/1564742781129502722).
+
+### Test Harnesses
+
+#### Internal Functions
+
+To test `internal` functions, write a harness contract that inherits from the contract under test (CuT). Harness contracts that inherit from the CuT expose the `internal` functions as `external` ones.
+
+Each `internal` function that is tested should be exposed via an external one with a name that follows the pattern `exposed_<function_name>`. For example:
+
+```solidity
+// file: src/MyContract.sol
+contract MyContract {
+  function myInternalMethod() internal returns (uint) {
+    return 42;
+  }
+}
+
+// file: test/MyContract.t.sol
+import {MyContract} from "src/MyContract.sol";
+
+contract MyContractHarness is MyContract {
+  // Deploy this contract then cll this method to test `myInternalMethod`.
+  function exposed_myInternalMethod() external returns (uint) {
+    return myInternalMethod();
+  }
+}
+```
+
+#### Private Functions
+
+Unfortunately there is currently no good way to unit test `private` methods since they cannot be accessed by any other contracts. Options include:
+
+- Converting `private` functions to `internal`.
+- Copy/pasting the logic into your test contract and writing a script that runs in CI check to ensure both functions are identical.
+
+#### Workaround Functions
+
+Harnesses can also be used to expose functionality or information otherwise unavailable in the original smart contract. The most straightforward example is when we want to test the length of a public array. The functions should follow the pattern: `workaround_<function_name>`, such as `workaround_queueLength()`.
+
+Another use case for this is tracking data that you would not track in production to help test invariants. For example, you might store a list of all token holders to simplify validation of the invariant "sum of all balances must equal total supply". These are often known as "ghost variables". You can learn more about this in [Rikard Hjort](https://twitter.com/rikardhjort)'s [Formal Methods for the Working DeFi Dev](https://youtu.be/TiuEWMo6w8U?t=3142) talk.
+
+### Best practices
+
+Thanks to [@samsczun](https://twitter.com/samczsun)'s [How Do You Even Write Secure Code Anyways](https://www.youtube.com/watch?v=Wm3t8Fuiy1E) talk for the tips in this section and the following section.
+
+- Don't optimize for coverage, optimize for well thought-out tests.
+- Write _positive_ unit tests for things that the code should handle. Validate _all_ state that changes from these tests.
+- Write _negative_ unit tests for things that the code should _not_ handle. It's helpful to follow up (as an adjacent test) with the positive test and make the change that it needs to pass.
+- Write integration tests to test entire features.
+- Write fork tests to verify the correct behavior with existing deployed contract.
+
+### Taint Analysis
+
+When testing, you should prioritize functions that an attacker can affect, that means functions that accept some kind of user input. These are called _sources_.
+
+Consider that input data as _tainted_ until it has been checked by the code, at which point it's considered _clean_.
+
+A _sink_ is a part of the code where some important operation is happening. For example, in MakerDAO that would be `vat.sol`.
+
+You should _ensure_ that no _tainted_ data ever reaches a _sink_. That means that all data that find themselves in the sink, should, at some point, have been checked by you. So, you need to define what the data _should_ be and then make sure your checks _ensure_ that the data will be how you expect it to be.
+
+## Scripts
+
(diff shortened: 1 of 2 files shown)


odyslam:
@mds1 I left some replies in the comments that were not resolved. Did you address them (include/not include) or should I move them here.

mds1:
Yea let's move them here, I don't think all were addressed, thanks!

src/tutorials/best-practices.md
```diff
@@ -0,0 +1,200 @@
...
+1. For unit tests, treat contracts as `describe` blocks:
+
+   - `contract Add` holds all unit tests for the `add` method.
+   - `contract Supply` holds all tests for the `supply` method.
+   - `contract Constructor` hold all tests for the constructor.
+   - One benefit of this approach is that smaller contracts should compile faster than large ones, so this approach of many small contracts should save time as test suites get large.
+
+1. Integration tests should live in the same `test` directory, with a clear naming convention. These may be in dedicated files, or they may live next to related unit tests in existing test files.
+
+1. Be consistent with test naming, as it's helpful for filtering tests (e.g. for gas reports you might want to filter out revert tests). When combining naming conventions, keep them alphabetical.
```
odyslam, line 67: Lucas suggestion: `test_functionName_description`

I am in favour

src/tutorials/best-practices.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Best practices
+
+Thanks to [@samsczun](https://twitter.com/samczsun)'s [How Do You Even Write Secure Code Anyways](https://www.youtube.com/watch?v=Wm3t8Fuiy1E) talk for the tips in this section and the following section.
+
+- Don't optimize for coverage, optimize for well thought-out tests.
+- Write _positive_ unit tests for things that the code should handle. Validate _all_ state that changes from these tests.
+- Write _negative_ unit tests for things that the code should _not_ handle. It's helpful to follow up (as an adjacent test) with the positive test and make the change that it needs to pass.
+- Write integration tests to test entire features.
+- Write fork tests to verify the correct behavior with existing deployed contract.
```
odyslam, line 135: Where should fork tests live?

IMHO, they should not be run as often as unit tests, as they would slow down the developer.

odyslam (changes requested):

src/tutorials/best-practices.md
```diff
@@ -0,0 +1,200 @@
...
+    - [Workaround Functions](#workaround-functions)
+  - [Best practices](#best-practices)
+  - [Taint Analysis](#taint-analysis)
+- [Scripts](#scripts)
+- [Comments](#comments)
+- [Resources](#resources)
+
+## General Contract Guidance
+
+1. Always use "absolute" paths for imports, i.e. paths relative to the repo root. This makes it easier to see where files are from and reduces churn when moving files around.
```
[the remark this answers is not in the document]
lucas-manuel: `import { MapleLoan as Loan } from "../modules/loan/contracts/MapleLoan.sol";`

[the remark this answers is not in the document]
lucas-manuel: This was something we adopted a while back because we kept having conflicting submodule versions, so this was a way around that. Also natively works with the solc static compiler in VSCode without settings updates but thats not a big deal

src/tutorials/best-practices.md
lucas-manuel: I like this pattern a lot because it makes searching for tests super easy. I do get @mds1s point around integration tests, but maybe this could be the unit test convention?

src/tutorials/best-practices.md
```diff
@@ -0,0 +1,200 @@
...
+   - `testFuzz_Description` for fuzz tests.
+   - `test_Revert[If|When]_Condition` for tests expecting a revert.
+   - `testFork_Description` for tests that fork from a network.
+   - `testForkFuzz_Revert[If|When]_Condition` for a fuzz test that forks and expects a revert.
+
+1. Be careful with with fuzz tests on a fork to avoid burning through RPC requests with non-deterministic fuzzing.
+
+1. When using assertions like `assertEq`, consider leveraging the last string param to make it easier to identify failures. These can be kept brief, or even just be numbers&mdash;they basically serve as a replacement for showing line numbers of the revert, e.g. `assertEq(x, y, "1")` or `assertEq(x, y, "sum1")`. _(Note: [Foundry#2328](https://github.com/foundry-rs/foundry/issues/2328) tracks integrating this natively)._
+
+1. Don't forget invariant tests! For the assertion string, use a verbose english description of the invariant: `assertEq(x + y, z, "Invariant violated: the sum of x and y must always equal z")`.
```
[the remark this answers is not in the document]
lucas-manuel: Thinking about talking about actor based patterns etc here

src/tutorials/best-practices.md
```diff
@@ -0,0 +1,200 @@
...
+
+Another use case for this is tracking data that you would not track in production to help test invariants. For example, you might store a list of all token holders to simplify validation of the invariant "sum of all balances must equal total supply". These are often known as "ghost variables". You can learn more about this in [Rikard Hjort](https://twitter.com/rikardhjort)'s [Formal Methods for the Working DeFi Dev](https://youtu.be/TiuEWMo6w8U?t=3142) talk.
+
+### Best practices
+
+Thanks to [@samsczun](https://twitter.com/samczsun)'s [How Do You Even Write Secure Code Anyways](https://www.youtube.com/watch?v=Wm3t8Fuiy1E) talk for the tips in this section and the following section.
+
+- Don't optimize for coverage, optimize for well thought-out tests.
+- Write _positive_ unit tests for things that the code should handle. Validate _all_ state that changes from these tests.
```
lucas-manuel, line 132: Maybe we should explicitly mention that each code path should have a separate unit test?

src/tutorials/best-practices.md
```diff
@@ -0,0 +1,200 @@
...
+
+   - Define `[rpc_endpoints]` in the config file and use the `createSelectFork` family of cheatcodes.
+   - Always pin to a block so tests are deterministic and RPC responses are cached.
+   - More info on this fork test approach [here](https://twitter.com/msolomon44/status/1564742781129502722).
+
+### Test Harnesses
+
+#### Internal Functions
+
+To test `internal` functions, write a harness contract that inherits from the contract under test (CuT). Harness contracts that inherit from the CuT expose the `internal` functions as `external` ones.
```
lucas-manuel, line 91: Idk where this would go, but one cool thing we've been doing is using contracts (idk whether to refer to them as harnesses or mocks in this context) in integration fuzzing tests to allow for easier fuzzing. An example is a totalAssets setter that is added to a contract that inherits a 4626 implementation.

`vm.etch` can be used to replace this in the broader context of a protocol and fuzzing can be done more easily/effectively in an integration test this way

src/tutorials/best-practices.md
```diff
@@ -0,0 +1,200 @@
...
+- Delete unused code.
+
+## Tests
+
+1. For testing `MyContract.sol`, the test file should be `MyContract.t.sol`. For testing `MyScript.s.sol`, the test file should be `MyScript.t.sol`.
+
+   - If the contract is big and you want to split it over multiple files, just group them logically like `MyContract.owner.t.sol`, `MyContract.deposits.t.sol`, etc.
+
+1. Never make assertions in the `setUp` function, instead use a dedicated test like `test_SetUpState()` if you need to ensure your `setUp` function does what you expected. More info on why in [Foundry#1291](https://github.com/foundry-rs/foundry/issues/1291)
+
```
lucas-manuel, line 57: Should we add a section on events assertions?

src/tutorials/best-practices.md
```diff
@@ -0,0 +1,200 @@
...
+- Use descriptive variable names.
+- Limit the number of active variables.
+- No redundant logic.
+- Early exit as much as possible to reduce mental load when seeing the code.
+- Related code should be placed near each other.
+- Delete unused code.
+
+## Tests
+
+1. For testing `MyContract.sol`, the test file should be `MyContract.t.sol`. For testing `MyScript.s.sol`, the test file should be `MyScript.t.sol`.
```
lucas-manuel, line 52: Also we should fix this list as they are currently all `1.`

src/tutorials/best-practices.md
```diff
@@ -0,0 +1,200 @@
...
+Unfortunately there is currently no good way to unit test `private` methods since they cannot be accessed by any other contracts. Options include:
+
+- Converting `private` functions to `internal`.
+- Copy/pasting the logic into your test contract and writing a script that runs in CI check to ensure both functions are identical.
+
+#### Workaround Functions
+
+Harnesses can also be used to expose functionality or information otherwise unavailable in the original smart contract. The most straightforward example is when we want to test the length of a public array. The functions should follow the pattern: `workaround_<function_name>`, such as `workaround_queueLength()`.
+
+Another use case for this is tracking data that you would not track in production to help test invariants. For example, you might store a list of all token holders to simplify validation of the invariant "sum of all balances must equal total supply". These are often known as "ghost variables". You can learn more about this in [Rikard Hjort](https://twitter.com/rikardhjort)'s [Formal Methods for the Working DeFi Dev](https://youtu.be/TiuEWMo6w8U?t=3142) talk.
```
lucas-manuel, line 125: Yeah I think we should add a dedicated section about ghost variables to the invariant documentation

lucas-manuel: We can also discuss some of the other concepts from that talk (naive implementations etc)

src/tutorials/best-practices.md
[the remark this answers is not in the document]
mds1: We just use the default remappings most of the time, occasionally may add others. My version @lucas-manuel's example would be something like:

```solidity
import { MapleLoan as Loan } from "src/loan/contracts/MapleLoan.sol";
```

However, @mattsse has mentioned these absolute paths can cause issues sometimes, and I've also noticed slither can't handle them. Even though slither should handle them, maybe we should remove this line since they are sometimes troublesome?

src/tutorials/best-practices.md
mds1: What's the rationale for this convention (`test_functionName_description`) vs. "contract name = function under test, test names are spec of that function" as currently proposed? The logic I have for the current approach is that in JS/TS tests with hardhat, structuring contract unit tests such that describe blocks are functions, and `it('whatever')` is a requirement of that function is common, e.g. see [Uni V3 tests](https://github.com/Uniswap/v3-core/blob/main/test/UniswapV3Pool.spec.ts#L159). This makes that transition seamless.

It still makes searching for tests easy: just search for `contract FunctionName` instead of `test_functionName`

Without it, what becomes the "describe" block for a function? I guess you can create arbitrary scopes with comments, but that feels a bit grosser

src/tutorials/best-practices.md
mds1: These days I actually feel pretty strongly that fork tests should not get special treatment, and should be used liberally. Here's why:

1. Mocks are arguably a remnant of closed-source web2 dev—you have to mock API responses because the code for that API isn't open source so you can't just run it locally. But for blockchains that's not true: any code you're interacting with that's already deployed can be locally executed and even modified for free. So why introduce the risk of a wrong mock if you don't need to?

2. A big reason to avoid forking and use mocks is that fork tests are slow. But this is not true. By pinning to a block number, RPC responses are cached and only the first run is slower, subsequent runs should be very fast (see https://github.com/mds1/convex-shutdown-simulation/ as proof, 7 minutes for remote RPC, 13s for a local RPC, half a second with cached results). Alchemy and Infura both offer free archive data, so pinning to a block shouldn't be problematic either.

src/tutorials/best-practices.md
mds1: I like that, just added 888a768 with this info

src/tutorials/best-practices.md
mds1: > An example is a totalAssets setter that is added to a contract that inherits a 4626 implementation.

The way we've done this is to add test helper methods like `function supply(address from, uint256 amount)` that deals token to `from`, then uses prank to execute approvals and supply tokens to the protocol to increase `totalAssets`. Both approaches are definitely valid and have different tradeoffs / depends on the specifics of your protocol which is best. 

I think this could fall under the `Workaround Function` header if you wanted to add a few sentences

src/tutorials/best-practices.md
odyslam: What tips do you have in mind?

mds1: What exactly do you have in mind here? One thing I can think of is always using `vm.expectEmit(true, true, true, true)` unless you have a strong reason to use `false` somewhere. Because:
1. This ensures you test everything in your event.
2. If you add a topic (i.e. a new indexed param), it's now tested by default.
3. Even if you only have 1 topic, the extra `true` args don't hurt

src/tutorials/best-practices.md
mds1: Heh I actually did that on purpose because markdown automatically renders the numbers sequentially, so using `1.` everywhere makes it easier to reorder/add/remove things without the numbers getting out of sync

src/tutorials/best-practices.md
mds1: I agree, I thought there were some great topics in that talk and am down to document them. Should just expand this "workarounds" section to include more info? (Feel free to push a commit with this info, or if you don't have permission can open a PR into this branch)

src/tutorials/best-practices.md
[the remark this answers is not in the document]
odyslam: I started using full paths, and I have to say that it looks much cleaner than mentally having to work out the `../../` and mentally count how many times you have to go back, etc etc.

That being said, other tools do have an issue with those.

I really don't know what to suggest.

src/tutorials/best-practices.md
odyslam: Okay, agree here. 

Let's do that

src/tutorials/best-practices.md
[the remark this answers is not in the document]
odyslam: Agreed with mds1

src/tutorials/best-practices.md
odyslam: I have sort of added ad hoc helper functions inside the test contract. If you mean helper functions in the Contract Under Test (e.h harness), then we should add it as a `workaround_` which is to signify that the function is not normally available to the contract but is added to the harness/mock for utility. 

What do you think @lucas-manuel

src/tutorials/best-practices.md
odyslam: Perhaps also add an `address` field to verify the event's origin? TBH I don't do it always cause "ahem", reasons, but I think it should be a best practice.

@lucas-manuel @mds1 

Can't wait for negative assertions (expectNotEmit).

odyslam:
Okay, @mds1 @lucas-manuel addressed all your comments. I might have bw for the last (Rikard's suggestions).

Can we resolve all that remain and move to ship this? I think it's **excellent** work and tbh I am super excited to get some feedback 🔥

odyslam:
Should we add something about code coverage? How accurate is it? I haven't used it at all.

src/tutorials/best-practices.md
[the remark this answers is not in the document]
mds1: There's recently been other reports where using absolute paths breaks intellisense/autocomplete type of IDE features, e.g. https://github.com/foundry-rs/forge-std/pull/206

But yea, I agree with you. Maybe we just document these tradeoffs and devs can make their own decisions?

src/tutorials/best-practices.md
mds1: > I like being able to test a contract by just adding the --mc argument.

You can still do that, but with just the `--mp` flag instead, e.g. `forge test --mp test/MyContract.t.sol` 😛

> The "describe" is the `test_functionName` that you can visually observe in the final trace, as they will be grouped.

That's true, I think at the end of the day it's largely similar. You can argue that now then you lose a level of hierarchy/flexibility as now your file (`MyContract.t.sol`) only has a single contract (`contract MyContractTest`), so `--mc MyContractTest` and `--mp test/MyContract.t.sol` are equivalent. 

Granted, with this naming convention, I don't think you lose anything since you can still use the `--mt` flag to run all tests for a given function, though I guess you need to be a bit more clever with your regexes to match everything since some are `test_function` and others are `testFuzz_function`

I'm pretty sure the main benefit of the currently documented approach is that small contracts will compile faster, but tbh we probably need to benchmark this to confirm as I'm not 100% confident

> I would suggest: Let's keep the best practice in place and perhaps revise it. TBH, I need to test it extensively before I am comfortable making a suggestion other than what I have used and I like.

This is fair, I'm onboard with this path forward. I think this can be a living document that continually changes, because eth dev evolves fast

src/tutorials/best-practices.md
mds1: Should we add this info to the doc? I think your previous stance around fork tests is common, and we should convert people 😁

src/tutorials/best-practices.md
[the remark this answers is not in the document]
mds1: Cool, going to mark this resolved then

src/tutorials/best-practices.md
mds1: +1 on negative assertions, that will be useful

> Perhaps also add an `address` field to verify the event's origin

Is this supported? I thought it wasn't yet, and there's an issue here https://github.com/foundry-rs/foundry/issues/2921

mds1:
Hmm I don't know, I still see a lot of issues in the repo around coverage so maybe we leave it out for this first version, and add it in the future

src/tutorials/best-practices.md
mds1: (I think we can consider this resolved from a merge standpoint, and can leave it open to continue discussion if we want)

mds1:
Going to leave some comments later today to:
1. Add info per some of the outstanding comments above
2. Mention one more gotcha around scripting (a frontrunning issue, described in https://github.com/foundry-rs/foundry/issues/3589)

Then I'll take it out of draft for final review + merge

src/tutorials/best-practices.md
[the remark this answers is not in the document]
mds1: Switched to documenting tradeoffs in 935affb

src/tutorials/best-practices.md
mds1: Added to doc in 935affb

src/tutorials/best-practices.md
mds1: Added info on event testing in 935affb

mds1:
Ok added everything in the last two commits, this is ready for final review + merge

src/tutorials/best-practices.md
odyslam: @mds1, then why we can pass an address here? 
https://github.com/foundry-rs/forge-std/blob/ff4bf7db008d096ea5a657f2c20516182252a3ed/src/Vm.sol#L182

odyslam:
@mds1 let's ship this. I have added it to my backlog to view the video sometime in the next few days. Will take notes and revert back.

mds1:
Sounds good, deferring to @gakonst @onbjerg @mattsse to review/merge from here

src/tutorials/best-practices.md
odyslam: @mds1 I read the source code and indeed it checks for address. I will make a PR

gakonst:
Really appreciate you guys taking point here. Merged, and let's iterate on it separately.

Issue #687, opened 2022-10-31
Error when compiling NFT tutorial code


thomasbach-dev:
Taking the code from the “Implementing a basic NFT” section (https://book.getfoundry.sh/tutorials/solmate-nft#implement-a-basic-nft) in the NFT Tutorial, I get a compilation error:

```
 forge build
[⠒] Compiling...
[⠆] Compiling 5 files with 0.8.13
[⠰] Solc 0.8.13 finished in 23.37ms
Error: 
Compiler run failed
error[6275]: ParserError: Source "lib/openzeppelin-contracts/contracts/contracts/utils/Strings.sol" not found: File not found. Searched the following locations: "PATH/TO/REPOBASE/foundry-playground".
 --> src/NFT.sol:5:1:
  |
5 | import "openzeppelin-contracts/contracts/utils/Strings.sol";
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I installed the library and can find the following path: `lib/openzeppelin-contracts/contracts/utils/Strings.sol`.

Sabnock01:
Did you put the remappings in `foundry.toml` or `ramappings.txt`?

Sabnock01:
And how did you install the library? Through `forge install` or `npm`?

thomasbach-dev:
No, there was no remapping. So I just executed
> forge re > remappings.txt

The error stays the same.

> Did you put the remappings in `foundry.toml` or `ramappings.txt`?

thomasbach-dev:
> And how did you install the library? Through `forge install` or `npm`?

Using `forge install transmissions11/solmate Openzeppelin/openzeppelin-contracts`.

thomasbach-dev:
btw: current code can be found here: https://github.com/thomasbach-dev/foundry-playground/tree/nft-with-solmate

I used nix to install foundry via https://github.com/shazow/foundry.nix

mattsse:
this looks like an issue with the remappings auto-detection that wrongly concatenates `contracts` twice.

you should be able to fix this by adding a oz remappings directly into foundry.toml

thomasbach-dev:
Nice, I was indeed able to get the build running by adding
> remappings = ['openzeppelin-contracts/=lib/openzeppelin-contracts/']

to `foundry.toml`.

PR #689, opened 2022-11-05
Add instructions for remapping

thomasbach-dev:
I ran into this problem described in issue #687. As the tutorial should be as beginner friendly as possible it should really take the user by the hand and tell her what to do.


mattsse (approved):
sorry for the wait, tysm!

PR #693, opened 2022-11-09
Added example expectRever when using a four letter error message.

dejx:
Added example expectRever when using a four letter error message.


mattsse (approved):
lgtm

PR #696, opened 2022-11-14
Added other explorers in deploying.md

leviska:
`forge verify-contract` supports 3 ways of verifying the contract, but in `deploying.md` only Etherscan is mentioned.
I personally initially thought that `forge` supports only etherscan, but then I went to `verify-contract` docs, where I could find, that there are also 2 more verifiers supported


mattsse (approved):
nice, tysm!

PR #698, opened 2022-11-17
added example of encoding a struct/tuple/custom type with components

PatrickAlphaC:
Per https://github.com/foundry-rs/foundry/issues/3706

Just to add more docs information.


mattsse (approved):
ty, ser!

PR #704, opened 2022-11-21
Added Rust.exe and VS Code Installer for Windows

johnfawole:
I use Windows, and I encountered this installation issue. Windows users who do not have the VS Code Installer would not be able to successfully install Foundry. I just fixed that in this edit.


mattsse (approved):

johnfawole:
@all-contributors

PR #706, opened 2022-11-22
Update documentation for envString

paulrberg:
- [x] Explain what happens when the environment variable is undefined.
- [x] Mention that environment variables can be put in a `.env` file.


mattsse (approved):
ty!

PR #712, opened 2022-11-28
Expand on using fuzz tests with forks

webthethird:
The content of this PR is essentially taken verbatim from a response to my question on Telegram by @mds1, who gave me permission to PR this info into the book :)

### DIFF
--- a/src/tutorials/best-practices.md
+++ b/src/tutorials/best-practices.md
@@ -99,13 +99,12 @@ Additional best practices from [samsczun](https://twitter.com/samczsun)'s [How D
    - Mocks are _required_ in closed-source web2 development—you have to mock API responses because the code for that API isn't open source so you cannot just run it locally. But for blockchains that's not true: any code you're interacting with that's already deployed can be locally executed and even modified for free. So why introduce the risk of a wrong mock if you don't need to?
    - A common reason to avoid fork tests and prefer mocks is that fork tests are slow. But this is not always true. By pinning to a block number, forge caches RPC responses so only the first run is slower, and subsequent runs are significantly faster. See [this benchmark](https://github.com/mds1/convex-shutdown-simulation/), where it took forge 7 minutes for the first run with a remote RPC, but only half a second once data was cached results. Alchemy and Infura both offer free archive data, so pinning to a block shouldn't be problematic. (Note that you may need to configure your CI to cache the RPC responses between runs).
 
-1. Be careful with with fuzz tests on a fork to avoid burning through RPC requests with non-deterministic fuzzing.
+1. Be careful with with fuzz tests on a fork to avoid burning through RPC requests with non-deterministic fuzzing. If the input to your fork fuzz test is some parameter which is used in an RPC call to fetch data (e.g. querying the token balance of an address), each run of a fuzz test uses at least 1 RPC request, so you'll quickly hit rate limits or usage limits. Solutions to consider:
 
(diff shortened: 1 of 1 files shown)


src/tutorials/best-practices.md
```diff
@@ -101,6 +101,11 @@ Additional best practices from [samsczun](https://twitter.com/samczsun)'s [How D
 
 1. Be careful with with fuzz tests on a fork to avoid burning through RPC requests with non-deterministic fuzzing.
 
+    - If the input to your fork fuzz test is some parameter which is used in an RPC call to fetch data (e.g. maybe you query the token balance of an address), each run of a fuzz test uses 1+ RPC requests, so you'll quickly hit rate limits or usage limits.
+    - One solution is to specify a fuzz/invariant seed: this makes sure each forge test invocation uses the same fuzz inputs. RPC results are cached locally, so you'll only query the node the first time.
+    - Another solution is to structure your tests so the data you are fuzzing over is computed locally by your contract, and not data that is used in an RPC call (may or may not be feasible based on what you're doing).
+    - Lastly, you can of course always run a local node or bump your RPC plan.
+
```
mds1, line 108: Thanks! Small refactor suggestion, wdyt @webthethird? Also added in a reference to multicall + a placeholder where we should link to the docs

```suggestion
1. Be careful with with fuzz tests on a fork to avoid burning through RPC requests with non-deterministic fuzzing. If the input to your fork fuzz test is some parameter which is used in an RPC call to fetch data (e.g. querying the token balance of an address), each run of a fuzz test uses at least 1 RPC request, so you'll quickly hit rate limits or usage limits. Solutions to consider:

    - Replace multiple RPC calls with a single [multicall](https://github.com/mds1/multicall).
    - Specify a fuzz/invariant [seed](TODO add link to correct page in docs): this makes sure each `forge test` invocation uses the same fuzz inputs. RPC results are cached locally, so you'll only query the node the first time.
    - Structure your tests so the data you are fuzzing over is computed locally by your contract, and not data that is used in an RPC call (may or may not be feasible based on what you're doing).
    - Lastly, you can of course always run a local node or bump your RPC plan.
```

webthethird: Looks good to me!

mattsse (approved):
lgtm, pending @mds1

src/tutorials/best-practices.md
```diff
@@ -99,8 +99,12 @@ Additional best practices from [samsczun](https://twitter.com/samczsun)'s [How D
...
 
-1. Be careful with with fuzz tests on a fork to avoid burning through RPC requests with non-deterministic fuzzing.
+1. Be careful with with fuzz tests on a fork to avoid burning through RPC requests with non-deterministic fuzzing. If the input to your fork fuzz test is some parameter which is used in an RPC call to fetch data (e.g. querying the token balance of an address), each run of a fuzz test uses at least 1 RPC request, so you'll quickly hit rate limits or usage limits. Solutions to consider:
 
+    - Replace multiple RPC calls with a single [multicall](https://github.com/mds1/multicall).
+    - Specify a fuzz/invariant [seed](TODO add link to correct page in docs): this makes sure each `forge test` invocation uses the same fuzz inputs. RPC results are cached locally, so you'll only query the node the first time.
```
mds1, line 105: This should link to `src/reference/config/testing.md#seed` (double check that path/link), currently there's a TODO placeholder in there 🙂

webthethird: Done!

PR #714, opened 2022-11-29
Update solidity-scripting.md

haythem96:
vm.startBroadcast() take an address as an arg, not a private key


mattsse (commented):
there should be two overloaded `startBroadcast` for uint256 and address. so this step seems redundant

gakonst:
closing as matt said

PR #718, opened 2022-11-30
Source bashrc

baribari2:
Adds a note that informs users of the possible need to run `source ~/.bashrc` before using `foundryup`

### DIFF
--- a/src/getting-started/installation.md
+++ b/src/getting-started/installation.md
@@ -22,6 +22,11 @@ foundryup
 
 If everything goes well, you will now have three binaries at your disposal: `forge`, `cast`, and `anvil`.
 
+
+> ⚠️ **Note**
+>
+> You may have to run the command `source ~/.bashrc` in order to use `foundryup`.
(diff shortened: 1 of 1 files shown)


src/getting-started/installation.md
```diff
@@ -22,6 +22,11 @@ foundryup
 
 If everything goes well, you will now have three binaries at your disposal: `forge`, `cast`, and `anvil`.
 
+
+> ⚠️ **Note**
+>
+> You may have to run the command `source ~/.bashrc` in order to use `foundryup`.
```
mds1, line 28: This is shell specific, e.g. it's `source ~/.zshrc` if using zsh, and other shells will be different.

Also, when you run the install command, `curl -L https://foundry.paradigm.xyz | bash`, it prints this to the terminal:

```
Detected your preferred shell is <shell> and added foundryup to PATH. Run '<command>' or 
start a new terminal session to use foundryup.
```

PR #735, opened 2022-12-15
Improve documentation for "expectRevert"

paulrberg:
## Description

I tested with Solidity v0.8.17

1. The current argument name `msg` throws the following compiler warning:

> Warning: This declaration shadows a builtin symbol

2. `expectRevert("AAAA")` cannot be compiled at all:

> TypeError: Member "expectRevert" not unique after argument-dependent lookup in contract Vm.

3. There is no need to cast the string literals to `bytes()`. The compiler works just fine without the cast.
4. There is no need to wrap the string literals in `abi.encodePacked` in the final example.

## Changelog

- [x] Rename "msg" arg to "message" to avoid compiler warning
- [x] Explain why `expectRevert("AAAA")` doesn't work
- [x] Remove the "bytes()" casting and use string literals in "expectRevert" docs
- [x] Pass string literals instead of using "abi.encodePacked"
- [x] Add note about when message-less reverts happen

### DIFF
--- a/src/cheatcodes/expect-revert.md
+++ b/src/cheatcodes/expect-revert.md
@@ -7,16 +7,16 @@ function expectRevert() external;
 ```
 
 ```solidity
-function expectRevert(bytes4 msg) external;
+function expectRevert(bytes4 message) external;
 ```
 
 ```solidity
-function expectRevert(bytes calldata msg) external;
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this is awesome, ty

PR #737, opened 2022-12-22
Corrected 'cheats' to 'vm'

L0GYKAL:
The keyword 'cheats' doesn't work with the last version of Foundry, it should be replaced by 'vm' in the doc


mattsse (approved):
ty

PR #738, opened 2022-12-22
docs: add cast sig-event

stanleyjzheng:
Brings docs in line with https://github.com/foundry-rs/foundry/pull/3929


mattsse (approved):
awesome thanks again!

PR #740, opened 2022-12-23
Say "testCannot" instead of "testRevert" in Best Practices

paulrberg:
The Foundry Book is currently giving contradictory guidance:

In [Writing Tests](https://book.getfoundry.sh/forge/writing-tests?highlight=testCannot#writing-tests), we are told that:

> A good practice is to use something like `testCannot` in combination with the [`expectRevert`](https://book.getfoundry.sh/cheatcodes/expect-revert.html) cheatcode

But then, in [Best Practices](https://book.getfoundry.sh/tutorials/best-practices), we are told this:

> `test_Revert[If|When]_Condition` for tests expecting a revert.

This PR standardizes the use of `testCannot`, because it's a stronger check than `test_Revert`. `testCannot` ensure that Foundry will not pass the test if there is no revert, whereas `test_Revert` doesn't.


mds1:
I'm biased since I wrote this section, but personally I prefer `test_Revert over `testCannot` for a few reasons:
1. We're building a [tool](https://github.com/ScopeLift/scopelint) that relies on that convention, where the pre-underscore text references the test type, and after the first underscore is test details. (This does allow `test_Cannot` though). (Maybe one day we'll add a config file but no plans for that currently).
2. "Revert" is more explicit in what you're expecting to happen and maps directly to the code, where as "Cannot" doesn't imply a revert.

Maybe a better approach is to let people choose and suggest alternatives, instead of dictating one over the other here?

paulrberg:
Oh, [scopelift](https://github.com/ScopeLift/scopelint) looks dope! Can't wait for it to be ready for use, so I can swap out solhint with it in my projects.

Rest of the points - agree with everything. "Revert" is definitely more explicit. My misconception above was to think that `testCannot` will explicitly look for a revert, and fail the test if there is no revert, e.g.

```solidity
// I thought that this doesn't passes, but it does
function testCannotFoo() external {
    assertTrue(true);
}
```

Not sure if it would be worth it though to make `testCannot` and `test_Revert` behave like `testFail` (just thinking out loud).

mds1:
> Oh, [scopelift](https://github.com/ScopeLift/scopelint) looks dope! Can't wait for it to be ready for use, so I can swap out solhint with it in my projects.

Thanks! It does still need some work but we are using it on a few projects, so you can definitely start trying it now (`cargo install scopelint` then `scopelint fmt` or `scopelint check`, and can use it in CI like [this](https://github.com/ScopeLift/foundry-template/blob/070e804b5847a0f664d27c12f835207f326b9345/.github/workflows/ci.yml#L74-L98)) and opening some issues with feature requests or bugs 

> Not sure if it would be worth it though to make `testCannot` and `test_Revert` behave like `testFail` (just thinking out loud).

There's been some talk of deprecating `testFail` fwiw since it's considered an antipattern now. The context is that it's legacy from dapptools which did not have `vm.expectRevert`, so you had to use `testFail`. And with `testFail` you can't test for specific revert reasons with it, so tests might pass for the wrong reason and there's no way to know. People have also named tests things like `testFailedSwap` and the end up confused with unexpected results since that has the same `testFail` prefix

paulrberg:
> you can definitely start trying it now (cargo install scopelint then scopelint fmt or scopelint check, and can use it in CI like [this](https://github.com/ScopeLift/foundry-template/blob/070e804b5847a0f664d27c12f835207f326b9345/.github/workflows/ci.yml#L74-L98))

Nice, I'll take it for a spin.

> There's been some talk of deprecating `testFail` fwiw since it's considered an anti-pattern now

Ooh, I see. I agree that testing with `vm.expectRevert` is much better than `testFail`.

Issue #660, opened 2022-09-30
Document cast create2


iFrostizz:
document: https://github.com/foundry-rs/foundry/pull/2467

iFrostizz:
can you assign me please, I will do it if there is no argument against it.

mattsse:
closed via #744

PR #744, opened 2022-12-24
Document `cast create2`

### LINKED ISSUE #520: Document how `create2` works during scripting
Add an example to the Solidity Scripting section, and explain how it works under the hood.

ref https://github.com/foundry-rs/foundry/issues/2518#issuecomment-1212028294

Sabnock01:
Resolves #520


mattsse (approved):
ty

iFrostizz:
Shouldn't this issue close #660 instead ?
Also, looks like we could add some hints about the option flags and mention it in the "cast reference" as well

Sabnock01:
> 

Yes it should. I mixed up the issue #'s. :upside_down_face:

PR #751, opened 2023-01-03
Update Formatter guide with latest configs

paulrberg:
The [Formatter](https://book.getfoundry.sh/reference/config/formatter) guide is outdated. It references a key that no longer exists (`func_attrs_with_params_multiline`), which was replaced in https://github.com/foundry-rs/foundry/pull/3252 by `multiline_func_header`.

This PR updates the formatter guide to contain the most up-to-date configs, noting that I have sourced the English descriptions from the [`fmt.rs`](https://github.com/foundry-rs/foundry/blob/be7084e74247a9880c4357d444143d03fc2269e5/config/src/fmt.rs) file. For `ignore`, I have assumed that the globs are comma-separated, but I am not sure about that.


mattsse (approved):
tysm!

PR #754, opened 2023-01-05
Add high level overview to `solidity-scripting`

joshieDo:
Related to https://github.com/foundry-rs/foundry/issues/3988 and some other wrong expectations that sometimes I see around.

Tried to make a concise explanation on how `forge script` works and what to expect


joshieDo:
@CodeSandwich 

does this help, and what things would you like to be explained in a clearer way?

mattsse (approved):

PR #755, opened 2023-01-05
chore: update custom error syntax

mitche50:
Solidity imports custom errors globally - referencing MyContract.CustomError results in an error.


mattsse (approved):
cleaner

PR #758, opened 2023-01-06
fix: delete extra hyphen in anvil script

paulrberg:
The command is either `anvil --mnemonic` or `anvil -m`, never `anvil --m`.


mattsse (approved):
nice catch

Issue #497, opened 2022-08-03
Document invariant testing + patterns


gakonst:
https://github.com/foundry-rs/foundry/pull/1572

https://github.com/maple-labs/revenue-distribution-token/tree/e0eca03c2ff05c36000a097de678543d7234f7cc/contracts/test

document simple invariant breaking, the actor model from Maple's RDT, usage of targetContract(s) etc

Perelyn-sama:
Getting this done.

lucas-manuel:
Thoughts on patterns:

## Open Invariant Testing

This testing pattern deploys all relevant contracts in the `setUp` function and exposes all the functions in these contracts to the fuzzer. This pattern must use `revert_on_failure = false` in the foundry.toml, since this pattern will make calls with unbounded values.

| Pros | Cons |
| - | - |
| - Checks all functions<br>- Unbounded params | - Single actor<br>- Reversions |
  
## Bounded Actor-Based Invariant Testing

This testing pattern deploys all relevant contracts in the `setUp` function and deploys wrapper contracts to call the functions on the contract. These contracts act as actors in a more realistic simulation of interactions against the desired contracts. These contracts can have relevant permissions and will perform calls with parameters bounded to reasonable values. This pattern must use `revert_on_failure = true` in the foundry.toml, since this pattern should expose any unexpected reversions over the course of the fuzzing campaign.

User contract reference: https://github.com/maple-labs/revenue-distribution-token/blob/be9592fd72bfa7142a217507f2d5500a7856329e/contracts/test/accounts/Staker.sol#L46

To take it a step further, a contract can be used to dynamically deploy new actors during the fuzzing campaign and perform actions from these actors: https://github.com/maple-labs/revenue-distribution-token/blob/be9592fd72bfa7142a217507f2d5500a7856329e/contracts/test/accounts/Staker.sol#L97

Another use case for this pattern is to create a contract that will call the `vm.warp` cheatcode at random intervals to simulate the passage of time and ensure that invariants are still held. 

This pattern must add all non-wrapper contracts deployed in the `setUp` function into the `excludeContracts` array as to ensure that these contracts are only called through the wrapper contracts.

| Pros | Cons |
| - | - |
| - Many actors<br>- Can simulate passage of time<br>- Can expose unexpected reversions | - Potential human error in incorrectly implementations of wrapper functions |

## Unbounded Actor-Based Invariant Testing

[...]

PR #760, opened 2023-01-06
feat: Add invariant testing 

lucas-manuel:
closes https://github.com/foundry-rs/book/issues/497


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
gakonst, line 1: remove

projects/invariant_testing/Basic4626Deposit.sol
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    function deposit(uint256 assets_, address receiver_) external virtual override nonReentrant returns (uint256 shares_) {
+        _mint(shares_ = previewDeposit(assets_), assets_, receiver_, msg.sender);
+    }
+
+    function _mint(uint256 shares_, uint256 assets_, address receiver_, address caller_) internal {
+        require(receiver_ != address(0), "RDT:M:ZERO_RECEIVER");
+        require(shares_   != uint256(0), "RDT:M:ZERO_SHARES");
+        require(assets_   != uint256(0), "RDT:M:ZERO_ASSETS");
```
mds1, line 22: Needs `forge fmt` 😜

projects/invariant_testing/Basic4626Deposit.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: UNLICENSED
+// ANCHOR: all
```
mds1, line 2: What is `ANCHOR: all`?

projects/invariant_testing/Basic4626Deposit.sol
```diff
@@ -0,0 +1,36 @@
...
+pragma solidity 0.8.10;
+
+contract Basic4626Deposit {
+
+    uint256
```
mds1, line 7: I know this is a WIP, just reminder for us tomake sure the code compiles before merging

src/forge/advanced-testing.md
```diff
@@ -2,12 +2,12 @@
 
 Forge comes with a number of advanced testing methods:
 
-- [Fuzz Testing](./fuzz-testing.md)
 - [Differential Testing](./differential-ffi-testing.md)
+- [Fuzz Testing](./fuzz-testing.md)
+- [Invariant Testing](./invariant-testing.md)
```
mds1, line 7: Nit, I'd probably order these as fuzz > invariant > differential. Fuzz is simplest, invariant builds on it and is useful for most people, differential builds on fuzz too but is less commonly needed

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,109 @@
+# Invariant Testing
+## Overview
+
+Invariant testing allows for a set of mathematical invariants to be tested against randomized sequences of pre-defined function calls from pre-defined contracts. After each function call is performed, all defined invariants are asserted.
+
+Invariant testing is a powerful tool to expose incorrect logic in protocols. Due to the fact that function call sequences are randomized and have fuzzed inputs, invariant testing can expose false assumptions and incorrect logic in edge cases and highly complex protocol states.
+
+Invariant testing campaigns have two dimensions, `runs` and `depth`.
+- `runs`: Number of times that a sequence of function calls is generated and run.
+- `depth`: Number of function calls made in a given `run`. All defined invariants are asserted after each function call is made.
```
mds1, line 10: Let's mention how reverted calls are treated within the `depth` counter (I think they are included and still increase depth when set to not fail on reverts?)

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,109 @@
...
+└─ function2: 50% (25%)
+
+targetContract2: 50%
+├─ function1: 25% (12.5%)
+├─ function2: 25% (12.5%)
+├─ function3: 25% (12.5%)
+└─ function4: 25% (12.5%)
+```
+
+This is something to be mindful of when designing target contracts, as target contracts with less functions will have each function called more often due to this probability distribution.
```
mds1, line 34: We should really fix this, cc @gakonst https://github.com/foundry-rs/foundry/issues/2986

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,109 @@
...
+<tr>
+
+<td>Direct assertions</td>
+<td>Directly query a protocol smart contracts and assert values are expected.</td>
+<td>
+
+```solidity
+assertGe(
+    token.totalAssets(),
+    token.totalSupply())
```
mds1, line 59:
```suggestion
    token.totalSupply()
```

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,109 @@
...
+</td>
+
+</tr>
+</table>
+
+### Conditional Invariants
+
+Invariants must hold for the course over the course of a given fuzzing campaign, but that doesn't mean they must hold true in every situation. There is the possibility for certain invariants to be introduced/removed in a given scenario (e.g., during a liquidation). For this a dedicated testing contract should be used.
+
+It is not recommended to introduce conditional logic into invariant assertions because they have the possibility of introducing false positives because of an incorrect code path.
```
mds1, line 103: Can you expand on this, maybe with an example?

mds1 (commented):
Love it so far, just took a pass and left some small comments. Think I'll need to see it a bit more fleshed out before coming up with some naming ideas

projects/invariant_testing/Basic4626Deposit.sol
gakonst: This lets us reference the whole file via a template in the book

src/forge/invariant-testing.md
gakonst: ACK

projects/invariant_testing/Basic4626Deposit.sol
lucas-manuel: Yeah I'm gonna get rid of this file altogether actually

src/forge/advanced-testing.md
lucas-manuel: Sounds good will do

src/forge/invariant-testing.md
lucas-manuel: Yeah can mention that

src/forge/advanced-testing.md
PaulRBerg: I second @mds1's comment - to my shame, this is the first time I hear about differential testing.

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,109 @@
...
+)
+```
+</td>
+
+</tr>
+
+<tr>
+
+<td>Naive implementation assertions</td>
+<td>Query the a protocol smart contract function and compare it against a naive and typically highly gas-inefficient implementation of the same desired logic.</td>
```
PaulRBerg, line 85: Typo: "the a".

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,109 @@
...
+)
+```
+</td>
+
+</tr>
+</table>
+
+### Conditional Invariants
+
+Invariants must hold for the course over the course of a given fuzzing campaign, but that doesn't mean they must hold true in every situation. There is the possibility for certain invariants to be introduced/removed in a given scenario (e.g., during a liquidation). For this a dedicated testing contract should be used.
```
PaulRBerg, line 101: We should delete "for the course" here.

PaulRBerg:
I also read through the first draft of this, and I like it a lot - it's straight to the point.

What I think would be very much worth documenting, especially at the beginning of the Markdown document, is the fact that Forge considers all functions that start with `invariant_` an invariant test.

PaulRBerg:
@lucas-manuel, quick terminology question.

In August 2022, you referred in this [comment](https://github.com/foundry-rs/book/issues/497#issuecomment-1208473205) to the target contracts that encapsulate the calls to the real contracts that are to be invariant tested as "actors".

However, in this PR, it looks like you have changed the name to "handler" - am I correct that you now prefer this terminology over "actor"? If yes, why is it that Maple V2 uses [actors](https://github.com/maple-labs/maple-core-v2/tree/00f01ae7175885f8d49ac201a1c72465e320b2f6/tests/invariants/actors)?

lucas-manuel:
> @lucas-manuel, quick terminology question.
> 
> In August 2022, you referred in this [comment](https://github.com/foundry-rs/book/issues/497#issuecomment-1208473205) to the target contracts that encapsulate the calls to the real contracts that are to be invariant tested as "actors".
> 
> However, in this PR, it looks like you have changed the name to "handler" - am I correct that you now prefer this terminology over "actor"?

Thats right - we're moving to handlers since the term "actor" was used when we had to use one contract per user, before pranking was possible. Now that multiple actors can be used within a give contract, we determined that handler should be used instead.

src/forge/invariant-testing.md
lucas-manuel: Fixed thanks

PaulRBerg:
Thanks for confirming, @lucas-manuel.

Speaking of pranking multiple actors - the [`useRandomLp`](https://github.com/maple-labs/maple-core-v2/blob/00f01ae7175885f8d49ac201a1c72465e320b2f6/tests/invariants/actors/LpHandler.sol#L68-L73) modifier used in maple-core-v2 is something that all invariant users may need. I think that it would be worth it to rename it to something more general-purpose (e.g. `useRandomUser` or `useRandomActor`) and ship it with the `InvariantTest` contract part of Forge Std.

PaulRBerg:
Potentially dumb question - can `vm.assume` be used in invariant tests?

I see no reference to this cheatcode in Lucas' examples, but I'm not sure if this is because Maple uses a different constraining system (`constrictToRange`), or because fuzzing assumptions are not meant to be used in an invariant context. And if not, why? It might be worth it to explain this in the docs.

If I understand this correctly, under "Bounded Testing", the user is expected to bound the inputs to avoid reverts.

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,303 @@
...
+└─ function4: 25% (12.5%)
+```
+
+This is something to be mindful of when designing target contracts, as target contracts with less functions will have each function called more often due to this probability distribution.
+
+### Target Contract Setup
+
+Target contracts can be set up using the following three methods:
+1. Contracts that are deployed in the `setUp` function are automatically added to the set of target contracts.
+2. Contracts that are added to the the `targetContracts` array are added to the set of target contracts. (TODO: check if this is possible during invariant campaigns yet)
```
pcaversaccio, line 127: The current wording is rather confusing (at least for me) - can you please formulate it in a way that people understand that if `targetContracts` method is present on a contract, forge will use the array returned by that method **instead** (and not all contracts in the `setUp` function).

lucas-manuel:
Gonna work to push this through now, what is missing? @mds1 @gakonst @PaulRBerg @pcaversaccio

lucas-manuel:
> Potentially dumb question - can `vm.assume` be used in invariant tests?
> 
> I see no reference to this cheatcode in Lucas' examples, but I'm not sure if this is because Maple uses a different constraining system (`constrictToRange`), or because fuzzing assumptions are not meant to be used in an invariant context. And if not, why? It might be worth it to explain this in the docs.
> 
> If I understand this correctly, under "Bounded Testing", the user is expected to bound the inputs to avoid reverts.

Correct me if I'm wrong @mds1 but I believe `vm.assume` would work in the same way as in fuzz tests for this. Haven't used it personally though.

lucas-manuel:
> Thanks for confirming, @lucas-manuel.
> 
> Speaking of pranking multiple actors - the [`useRandomLp`](https://github.com/maple-labs/maple-core-v2/blob/00f01ae7175885f8d49ac201a1c72465e320b2f6/tests/invariants/actors/LpHandler.sol#L68-L73) modifier used in maple-core-v2 is something that all invariant users may need. We should consider renaming it to something more general-purpose (e.g. `useRandomUser` or `useRandomActor`) and ship it with the `InvariantTest` contract part of Forge Std.

This is an interesting idea, thoughts @mds1 ?

src/forge/invariant-testing.md
lucas-manuel: Is this the case? I haven't tested this. Updated the wording though let me know if that's better.

pcaversaccio: Well @msd1 told me that yesterday :) so maybe he should confirm as well.

pcaversaccio:
> Gonna work to push this through now, what is missing? @mds1 @gakonst @PaulRBerg @pcaversaccio 

Proper descriptions on _all_ the invariant helper functions like `targetSender` etc. would be very nice.

PaulRBerg:
> I believe `vm.assume` would work in the same way as in fuzz tests for this

It doesn't. I'm getting `Failed Assertion` errors if I have `vm.assume` statements in the invariant tests. If this is not intended behavior, should I open an issue in `forge-std`?

> Proper descriptions on all the invariant helper functions

Yes! A markdown table for this would be awesome.

> Gonna work to push this through now, what is missing?

The greatest conceptual leap for me were the handler contracts - I spent a few days looking at the `maple-core-v2` repository, and the [`LoanHandler`][lh] and the [`LpHandler`][lph] contracts were the most difficult to understand.

I see that you added more explanations and a few diagrams about them in the meantime (6 hrs ago), which are awesome. But I think that there are still a few other core concepts that it would be good to document, too:

1. How should actors be generated? This may depend upon the business logic of the smart contract under test, but we should be able to come up with a few rules of thumb, e.g. generate the actors in the constructor and store them in an array.
2. How should actors and other contract storage items be fuzzed? In `maple-core-v2`, you are using these seed arguments (`warpSeed_`, `borrowerIndexSeed_`), and it took me a while to understand their role. It would also be good to mention the need for a `useRandomActor` modifier - at least until Foundry offers a built-in solution for this.
3. Should handler contracts inherit from `Test`? Or `Utils`? If not, how can they get access to `vm`, `bound`, and so forth?

[lh]: https://github.com/maple-labs/maple-core-v2/blob/main/tests/invariants/actors/LoanHandler.sol
[lph]: https://github.com/maple-labs/maple-core-v2/blob/00f01ae7175885f8d49ac201a1c72465e320b2f6/tests/invariants/actors/LpHandler.sol

mds1:
> It doesn't. I'm getting `Failed Assertion` errors if I have `vm.assume` statements in the invariant tests. If this is not intended behavior, should I open an issue in `forge-std`?

@PaulRBerg Nice catch, just created https://github.com/foundry-rs/foundry/issues/4190 to track

> Speaking of pranking multiple actors - the [`useRandomLp`](https://github.com/maple-labs/maple-core-v2/blob/00f01ae7175885f8d49ac201a1c72465e320b2f6/tests/invariants/actors/LpHandler.sol#L68-L73) modifier used in maple-core-v2 is something that all invariant users may need. We should consider renaming it to something more general-purpose (e.g. `useRandomUser` or `useRandomActor`) and ship it with the `InvariantTest` contract part of Forge Std.
> 
> > This is an interesting idea, thoughts @mds1 ?

This is an interesting idea. More generally maybe we can add a template `InvariantHandler` contract to forge-std. @lucas-manuel @PaulRBerg if either of you have ideas on what this could look like, create an issue in forge-std and we can continue the conversation there.

> Gonna work to push this through now, what is missing? @mds1 @gakonst @PaulRBerg @pcaversaccio

Awesome, thanks! Later today I'll review what's currently here and reply with some thoughts

PaulRBerg:
> More generally maybe we can add a template InvariantHandler contract to forge-std.

I like this idea. Definitely one or two helpers for actor generation and management would be nice. I may open an issue for this later.

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,321 @@
+# Invariant Testing
+## Overview
+
+Invariant testing allows for a set of mathematical invariants to be tested against randomized sequences of pre-defined function calls from pre-defined contracts. After each function call is performed, all defined invariants are asserted.
+
+Invariant testing is a powerful tool to expose incorrect logic in protocols. Due to the fact that function call sequences are randomized and have fuzzed inputs, invariant testing can expose false assumptions and incorrect logic in edge cases and highly complex protocol states.
+
+Invariant testing campaigns have two dimensions, `runs` and `depth`.
+- `runs`: Number of times that a sequence of function calls is generated and run.
+- `depth`: Number of function calls made in a given `run`. All defined invariants are asserted after each function call is made. If a function call reverts, the `depth` counter still increments.
```
mds1, line 10: @PaulRBerg do you think this explanation is clear? I know there was some confusion in https://github.com/foundry-rs/foundry/issues/3005 around what a `run` is

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,321 @@
...
+
+Similar to how standard tests are run in Foundry by prefixing a function name with `test`, invariant tests are denoted by prefixing the function name with `invariant` (e.g., `function invariant_A()`).
+
+## Defining Invariants
+
+Invariants are mathematical expressions that should always hold true over the course of a fuzzing campaign. A good invariant testing suite should have as many invariants as possible, and can have different testing suites for different protocol states.
```
mds1, line 16: They can also be things like "only the owner can do X" which isn't mathematical
 ```suggestion
Invariants are expressions that should always hold true over the course of a fuzzing campaign. A good invariant testing suite should have as many invariants as possible, and can have different testing suites for different protocol states.
```

File: src/forge/invariant-testing.md [hunk 15]
```diff
@@ -0,0 +1,321 @@
...
+
+There are different ways to assert invariants:
+
+<table>
+<tr><th>Type</th><th>Explanation</th><th>Example</th></tr>
+
+<tr>
+
+<td>Direct assertions</td>
+<td>Directly query a protocol smart contracts and assert values are expected.</td>
```
mds1, line 26:
```suggestion
<td>Directly query a protocol's smart contracts and assert values are expected.</td>
```

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,321 @@
...
+    sumBalanceOf
+)
+```
+</td>
+
+</tr>
+
+<tr>
+
+<td>Naive implementation assertions</td>
```
mds1, line 57: Should we call this "Deoptimizing"? It matchings the RV talk's name and is shorter. Or we can write something like 
```suggestion
<td>Deoptimizing (Naive implementation assertions)</td>
```

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,321 @@
...
+└─ function4: 25% (12.5%)
+```
+
+This is something to be mindful of when designing target contracts, as target contracts with less functions will have each function called more often due to this probability distribution.
+
+### Target Contract Setup
+
+Target contracts can be set up using the following three methods:
+1. Contracts that are deployed in the `setUp` function are automatically added to the set of target contracts.
+2. Contracts that are added to the the `targetContracts` array are added to the set of target contracts and used instead of the contract deployed in the `setUp`.
```
mds1, line 129: Typo. Also are we certain this is true? Ref https://github.com/foundry-rs/foundry/issues/4187
```suggestion
2. Contracts that are added to the the `targetContracts` array are added to the set of target contracts and used instead of the contracts deployed in the `setUp`.
```

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,321 @@
...
+3. Contracts that are deployed in the `setUp` can be **removed** from the target contracts if they are added to the `excludeContracts` array.
```
mds1, line 130: So really, the way you'd use these are:
- Option A: Pair (1) and (3)
- Option B: Use (2)
- Option C: More rare, all 3, if you need to manually add some already-existing contracts from e.g. a fork or `deployCode` usage

If that's correct, would be nice to give those sample use cases

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,321 @@
...
+[PASS] invariant_B() (runs: 50, calls: 10000, reverts: 5533)
+```
+
+### Invariant Handlers
+
+For more complex and integrated protocols, more sophisticated target contract usage is required to achieve the desired results. For example, a ERC-4626 based contract that accepts deposits of another ERC-20 token:
+
+```solidity
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.7;
```
mds1, line 195: May as well use a more recent version for all the pragmas
```suggestion
pragma solidity ^0.8.17;
```

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,321 @@
...
+    }
+```
+
+This contract will provide the necessary setup before a function call is made in order to ensure it is successful.
+
+Building on this concept, Handlers can be used to develop more sophisticated invariant tests. With Open invariant testing, the tests run as shown in the diagram below, with random sequences of function calls being made to the protocol contracts directly with fuzzed parameters. This will cause reverts for more complex systems as outlined above.
+
+![Blank diagram](https://user-images.githubusercontent.com/44272939/214752968-5f0e7653-d52e-43e6-b453-cac935f5d97d.svg)
+
+By using Handler contracts and **excluding** all protocol contracts from the `targetContracts` array by using the `excludeContracts` function, all function calls made to protocol contracts can be made in a way that is governed by the Handler to ensure successful calls. This is outlined in the diagram below.
```
mds1, line 311: This sentence contradicts the sentence near https://github.com/foundry-rs/book/pull/760/files#r1088430050, i.e. if you use `targetContracts` do you need to also exclude the ones from setup?

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,321 @@
...
+
+With this layer between the fuzzer and the protocol, more powerful testing can be achieved.
+
+Within Handlers, "ghost variables" can be tracked across multiple function calls to add additional info for invariant tests. A good example of this is summing all of the `shares` that each LP owns after depositing into the ERC-4626 token as shown above, and using that in the invariant (`totalSupply == sumBalanceOf`).
+
+Another benefit is the ability to perform assertions on function calls as they are happening. An example is asserting the ERC-20 balance of the LP has decremented by `assets` during the `deposit` function call. In this way, handler functions are similar to fuzz tests because they can take in fuzzed inputs, perform state changes, and assert before/after state.
+
+In addition, with Handlers, input parameters can be bounded to reasonable expected values such that `fail_on_revert` can be set to `true`. This ensures that every function call that is being made by the fuzzer must be successful against the protocol in order to get tests to pass. This is very useful for visibility and confidence that the protocol is being tested in the desired way.
```
mds1, line 321: Let's explicitly call out and link to the `bound` method for this input shaping. And hopefully soon `vm.assume` support too https://github.com/foundry-rs/foundry/issues/4190

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,321 @@
...
+
+![Invariant Diagrams - Page 2](https://user-images.githubusercontent.com/44272939/214752977-053f60e6-c644-42a9-8cff-4fd85a2517ac.svg)
+
+With this layer between the fuzzer and the protocol, more powerful testing can be achieved.
+
+Within Handlers, "ghost variables" can be tracked across multiple function calls to add additional info for invariant tests. A good example of this is summing all of the `shares` that each LP owns after depositing into the ERC-4626 token as shown above, and using that in the invariant (`totalSupply == sumBalanceOf`).
+
+Another benefit is the ability to perform assertions on function calls as they are happening. An example is asserting the ERC-20 balance of the LP has decremented by `assets` during the `deposit` function call. In this way, handler functions are similar to fuzz tests because they can take in fuzzed inputs, perform state changes, and assert before/after state.
```
mds1, line 319: Demonstrating some patterns like `useRandomUser` and modifying block properties through handler methods would be great to include in this section

src/forge/invariant-testing.md
PaulRBerg: @lucas-manuel and @pcaversaccio, can confirm that by using `targetContract`, the default behavior of Forge to automatically add all contracts deployed in the `setUp` function is turned off.

pcaversaccio: 👍 Awesome - I also tested it and confirm as well - at least with the latest nightly build version.

src/forge/invariant-testing.md
PaulRBerg: Ye, this is awesome.

src/forge/invariant-testing.md
PaulRBerg: I agree with @mds1 that we shouldn't define invariants as exclusively "mathematical expressions", but I don't think that example he gave is fitting in this context.

"only the owner can do X" sounds like a *function-level* invariant to me, whereas Foundry seems to me dealing with *system-level* invariants with its `invariant_` tests (the distinction is nicely explained by Anish Naik [here](https://youtu.be/QofNQxW_K08?t=2840)).

A better example would be time-series types of invariants, e.g. nonces that should only go up incrementally (as discussed in https://github.com/foundry-rs/foundry/issues/3005).

src/forge/invariant-testing.md
PaulRBerg: Ye, we are sure now that `targetContracts` overrides the default behavior to add all contracts from `setUp`.

https://github.com/foundry-rs/foundry/issues/4187#issuecomment-1406356194

src/forge/invariant-testing.md
PaulRBerg: I would bump up Option B and make it the first option mentioned - based on the experience I had with invariant testing so far, it seems to me that this will be the most common way invariant contracts will be targeted.

src/forge/invariant-testing.md
PaulRBerg: Nope, you don't.

https://github.com/foundry-rs/foundry/issues/4187#issuecomment-1406356194

src/forge/invariant-testing.md
PaulRBerg: Example modifier:

```solidity
address[] public actors;
address internal currentActor;

modifier useActor(uint256 actorIndexSeed) {
    currentActor = actors[bound(actorIndexSeed, 0, actors.length - 1)];
    vm.startPrank(currentActor);
    _;
    vm.stopPrank();
}
```

src/forge/invariant-testing.md
pcaversaccio: based on my personal previous experience with dapptools and now Foundry I can share the impression from @PaulRBerg.

pcaversaccio:
Another input regarding potential examples: it would be good to showcase quickly how to use properly `targetSender` since it might very useful for certain invariant tests. A more sophisticated example where I had to use it for my Vyper contract [`BatchDistributor`](https://github.com/pcaversaccio/snekmate/blob/feat/fuzz-tests/test/utils/BatchDistributor.t.sol#L412):

```solidity
contract BatchDistributorInvariants is Test, InvariantTest {
    VyperDeployer private vyperDeployer = new VyperDeployer();

    IBatchDistributor private batchDistributor;
    BatchDistributorHandler private batchDistributorHandler;

    ERC20Mock private erc20Mock;

    function setUp() public {
        batchDistributor = IBatchDistributor(
            vyperDeployer.deployContract("src/utils/", "BatchDistributor")
        );
        address msgSender = makeAddr("msgSender");
        erc20Mock = new ERC20Mock(
            "MyToken",
            "MTKN",
            msgSender,
            type(uint256).max
        );
        batchDistributorHandler = new BatchDistributorHandler(
            batchDistributor,
            erc20Mock
        );
        targetContract(address(batchDistributorHandler));
        targetSender(msgSender);
    }

    function invariantNoEtherBalance() public {
        assertEq(address(batchDistributor).balance, 0);
    }

    function invariantNoTokenBalance() public {
        /**
         * @dev This invariant breaks when tokens are sent directly to `batchDistributor`
         * as part of `distribute_token`. However, this behaviour is acceptable.
         */
        assertEq(erc20Mock.balanceOf(address(batchDistributor)), 0);
    }
}
```

src/forge/invariant-testing.md
gakonst, line 16:
```suggestion
Invariants are conditions expressions that should always hold true over the course of a fuzzing campaign. A good invariant testing suite should have as many invariants as possible, and can have different testing suites for different protocol states. Examples of an invariant would be "the xy=k formula always holds" for Uniswap, or "the sum of all owner balances is equal to the total supply" for an ERC20.
```

src/forge/invariant-testing.md
gakonst: i like

src/forge/invariant-testing.md
gakonst: +1 I would recommend B more so

src/forge/invariant-testing.md
gakonst: let's add paul's modifier above^

gakonst (changes requested):
tiny changes, supportive of all the comments, let's ship this!

src/forge/invariant-testing.md
lucas-manuel: Changed this to `defined assertions` to be more general, it gets defined more below.

lucas-manuel: Updated to this

src/forge/invariant-testing.md
lucas-manuel: @PaulRBerg https://www.youtube.com/watch?v=ETlNhV9jYJw&t=747s

lucas-manuel: Update

src/forge/invariant-testing.md
lucas-manuel: Moved around, lmk if this makes sense/should be reworded.

src/forge/invariant-testing.md
lucas-manuel: Updated

src/forge/invariant-testing.md
lucas-manuel: Nice TIL! Updated locally

PaulRBerg:
Speaking of useful modifiers, something like this may also be worth documenting:

```solidity
mapping(string => uint256) public calls;
uint256 public totalCalls;

modifier instrument(string memory func) {
    calls[func]++;
    totalCalls++;
    _;
}
```

This can be applied to handler contract functions like this:

```solidity
function myFunc() external instrument("myFunc") {
    // ...
}
```

The purpose being to get a call summary (e.g. [here][1]).

However, I am not sure whether adding a note about the above is worth it given the intention to add native support for instrumentation in Foundry (https://github.com/foundry-rs/foundry/issues/3607).

[1]: https://github.com/lucas-manuel/invariant-examples/blob/d6b4f06d7ebe0bbaebcfd666fb00433d30ab28da/test/WrappedTests.t.sol#L87-L101

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    function invariant_B() external {
+        assertGe(foo.val1() + foo.val2(), foo.val1());
+    }
+
+}
+```
+
+This setup will call `foo.addToA()` and `foo.addToB()` with a 50%-50% probability distribution with fuzzed inputs. Inevitably, the inputs will start to cause overflows and the function calls will start reverting. Since the default configuration in invariant testing is `fail_on_revert = true`, this will not cause the tests to fail. The invariants will hold throughout the rest of the fuzzing campaign and the result is that the test will pass. The output will look something like this:
```
giovannidisiena, line 182: Excellent documentation @lucas-manuel, thank you! I believe the default configuration here is `fail_on_revert = false` as per the [config README](https://github.com/foundry-rs/foundry/blob/master/config/README.md?plain=1#L183) which makes more sense in the context of this sentence and the final handler section.

gakonst:
@lucas-manuel I think one final pass based on the above comments and we're good to merge as a first iteration

pcaversaccio:
Haven't checked again the content of this PR but apparently you thought it would be a good idea to rename `InvariantTest` but `StdInvariant` (https://github.com/foundry-rs/forge-std/commit/b4f121555729b3afb3c5ffccb62ff4b6e2818fd3). Broke all of my invariant tests in the CI lol - in any case if we somewhere use the old term `InvariantTest` this should be updated.

PaulRBerg:
@pcaversaccio you should set the `branch` to `v1` in your `.gitmodules` (https://github.com/foundry-rs/forge-std/issues/275).

Nonetheless, I agree with you that all references to `InvariantTest` should be changd to `SdInvariant`.

pcaversaccio:
@PaulRBerg yeah agreed - however in my particular case here I didn't want to pin to v1 since the development iterations where important to me (i.e. invariant testing) and therefore I always pulled directly the `main` branch every two days or so. That's why I immediately realised that something is broken. But generally the pinning strategy you're suggesting is also something I usually follow.

pcaversaccio:
Also, I quickly wanted to emphasise that people should know that `Test` inherits from it _by default_ now. Because if you do the following `MyContract is Test, StdInvariant {...}` it will not compile due to the linearisation patter by Solidity. However, folks will read the release of `forge-std` and will get confused:
image

lucas-manuel:
> Speaking of useful modifiers, something like this may also be worth documenting:
> 
> ```solidity
> mapping(string => uint256) public calls;
> uint256 public totalCalls;
> 
> modifier instrument(string memory func) {
>     calls[func]++;
>     totalCalls++;
>     _;
> }
> ```
> 
> This can be applied to handler contract functions like this:
> 
> ```solidity
> function myFunc() external instrument("myFunc") {
>     // ...
> }
> ```
> 
> The purpose being to get a call summary (e.g. [here](https://github.com/lucas-manuel/invariant-examples/blob/d6b4f06d7ebe0bbaebcfd666fb00433d30ab28da/test/WrappedTests.t.sol#L87-L101)).
> 
> However, I am not sure whether adding a note about the above is worth it given the intention to add native support for instrumentation in Foundry ([foundry-rs/foundry#3607](https://github.com/foundry-rs/foundry/issues/3607)).

I like this idea but yeah I think we should push for the native support, thoughts @mds1 @gakonst ?

mds1:
> However, folks will read the release of forge-std and will get confused

The screenshot is correct when using v1.3.0 of forge-std, when v1.4.0 is released the release notes will document the breaking change and new usage info.

I agree with updating the docs to reflect this upcoming usage / naming though, cc @lucas-manuel 

> I like this idea but yeah I think we should push for the native support, thoughts @mds1 @gakonst ?

Agreed native support is preferable. There may be higher priority fixes/features though so I do think a template handler contract / modifiers in forge-std would be useful in the meantime. I also like [this pattern](https://twitter.com/eth_call/status/1622321144756928512) that I recently came across from @horsefacts, and people can then implement their own getters to parse the calldata

lucas-manuel:
> > However, folks will read the release of forge-std and will get confused
> 
> The screenshot is correct when using v1.3.0 of forge-std, when v1.4.0 is released the release notes will document the breaking change and new usage info.
> 
> I agree with updating the docs to reflect this upcoming usage / naming though, cc @lucas-manuel
> 
> > I like this idea but yeah I think we should push for the native support, thoughts @mds1 @gakonst ?
> 
> Agreed native support is preferable. There may be higher priority fixes/features though so I do think a template handler contract / modifiers in forge-std would be useful in the meantime. I also like [this pattern](https://twitter.com/eth_call/status/1622321144756928512) that I recently came across from @horsefacts, and people can then implement their own getters to parse the calldata

Should I add these to the docs?

lucas-manuel:
Just adding the target definitions and will push for final review.

lucas-manuel:
Alright good for review 👍

horsefacts:
> horsefact's pattern is still not too common so I'd avoid including it

Agree, this is more an exploration than something I'd recommend right now.

src/forge/invariant-testing.md
```diff
@@ -13,17 +13,21 @@ Similar to how standard tests are run in Foundry by prefixing a function name wi
 
 ## Defining Invariants
 
-Invariants are mathematical expressions that should always hold true over the course of a fuzzing campaign. A good invariant testing suite should have as many invariants as possible, and can have different testing suites for different protocol states.
+Invariants are conditions expressions that should always hold true over the course of a fuzzing campaign. A good invariant testing suite should have as many invariants as possible, and can have different testing suites for different protocol states.
 
-There are different ways to assert invariants:
+Examples of invariants are:
+- *"The xy=k formula always holds"* for Uniswap
+- *"The sum of all owner balances is equal to the total supply"* for an ERC-20 token.
```
mds1, line 20: Nit, since "owner" sounds like a privileged actor 
```suggestion
- *"The sum of all user balances is equal to the total supply"* for an ERC-20 token.
```

src/forge/invariant-testing.md
mds1: Yea @lucas-manuel `false` is the default here

https://github.com/foundry-rs/foundry/blob/4718a390de2afc7372082d77afe7dc8d0e435215/config/src/invariant.rs#L31

lucas-manuel:
> One comment [here](https://github.com/foundry-rs/book/pull/760#discussion_r1097742845) we should fix, but otherwise this looks really great! Awesome job @lucas-manuel 🔥

Updated! 👍

mds1:
@gakonst @lucas-manuel anything blocking merge here?

think we might just need info around how `targetContracts` can change since it automatically adds contracts deployed during an invariant run, cc @joshieDo for details

lucas-manuel:
> @gakonst @lucas-manuel anything blocking merge here?
> 
> think we might just need info around how `targetContracts` can change since it automatically adds contracts deployed during an invariant run, cc @joshieDo for details

Good to go on my end

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,475 @@
...
+
+### Invariant Test Helper Functions
+Invariant test helper functions are included in [`forge-std`](https://github.com/foundry-rs/forge-std/blob/master/src/StdInvariant.sol) to allow for configurable invariant test setup. The helper functions are outlined below:
+
+| Function | Description |
+|-|-|
+| `excludeContract(address newExcludedContract_)` | Adds a given address to the `_excludedContracts` array. This set of contracts is explicitly excluded from the target contracts.|
+| `excludeSender(address newExcludedSender_)` | Adds a given address to the to the `_excludedSenders` array. This set of addresses is explictly excluded from the target senders. |
+| `excludeArtifact(string memory newExcludedArtifact_)` | Adds a given string to the to the `_excludedArtifacts` array. This set of strings is explictly excluded from the target artifacts. |
+| `targetArtifact(string memory newTargetedArtifact_)` | Adds a given string to the to the `_targetedArtifacts` array. This set of strings is used for the target artifacts.  |
```
PaulRBerg, line 149: Double `to the to the` typo.

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,475 @@
...
+| `targetArtifactSelector(FuzzSelector memory newTargetedArtifactSelector_)` | Adds a given `FuzzSelector` to the to `_targetedArtifactSelectors` array. This set of `FuzzSelector`s is used for the target artifact selectors. |
+| `targetContract(address newTargetedContract_)` | Adds a given address to the to `_targetedContracts` array. This set of addresses is used for the target contracts. This array overwrites the set of contracts that was deployed during the `setUp`. |
+| `targetSelector(FuzzSelector memory newTargetedSelector_)` | Adds a given `FuzzSelector` to the to `_targetedSelectors` array. This set of `FuzzSelector`s is used for the target contract selectors. |
+| `targetSender(address newTargetedSender_)` | Adds a given address to the to `_targetedSenders` array. This set of addresses is used for the target senders. |
+
+
+### Target Contract Setup
+
+Target contracts can be set up using the following three methods:
+1. Contracts that are manually added to the the `targetContracts` array are added to the set of target contracts.
```
PaulRBerg, line 159: Double `the the` typo.

PaulRBerg (changes requested):
Just had a final read, looks lit. Thanks again @lucas-manuel for writing this up. There are only a few minor typos - besides these, LGTM.

I'm glad to see the `useActor` modifier included in this PR 🙌

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,475 @@
+# Invariant Testing
+## Overview
+
+Invariant testing allows for a set of invariant expressions to be tested against randomized sequences of pre-defined function calls from pre-defined contracts. After each function call is performed, all defined invariants are asserted.
+
+Invariant testing is a powerful tool to expose incorrect logic in protocols. Due to the fact that function call sequences are randomized and have fuzzed inputs, invariant testing can expose false assumptions and incorrect logic in edge cases and highly complex protocol states.
+
+Invariant testing campaigns have two dimensions, `runs` and `depth`.
```
pcaversaccio, line 8:
```suggestion
Invariant testing campaigns have two dimensions, `runs` and `depth`:
```

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,475 @@
...
+
+This is something to be mindful of when designing target contracts, as target contracts with less functions will have each function called more often due to this probability distribution.
+
+### Invariant Test Helper Functions
+Invariant test helper functions are included in [`forge-std`](https://github.com/foundry-rs/forge-std/blob/master/src/StdInvariant.sol) to allow for configurable invariant test setup. The helper functions are outlined below:
+
+| Function | Description |
+|-|-|
+| `excludeContract(address newExcludedContract_)` | Adds a given address to the `_excludedContracts` array. This set of contracts is explicitly excluded from the target contracts.|
+| `excludeSender(address newExcludedSender_)` | Adds a given address to the to the `_excludedSenders` array. This set of addresses is explictly excluded from the target senders. |
```
pcaversaccio, line 147:
```suggestion
| `excludeSender(address newExcludedSender_)` | Adds a given address to the to the `_excludedSenders` array. This set of addresses is explicitly excluded from the target senders. |
```

src/forge/invariant-testing.md
```diff
@@ -0,0 +1,475 @@
...
+| `excludeArtifact(string memory newExcludedArtifact_)` | Adds a given string to the to the `_excludedArtifacts` array. This set of strings is explictly excluded from the target artifacts. |
```
pcaversaccio, line 148:
```suggestion
| `excludeArtifact(string memory newExcludedArtifact_)` | Adds a given string to the to the `_excludedArtifacts` array. This set of strings is explicitly excluded from the target artifacts. |
```

pcaversaccio (commented):
Fixed typo: explictly ==> explicitly & I think `:` makes more sense where I suggested it. Anything else LGTM.

image

lucas-manuel:
Made all requested changes thanks @PaulRBerg @pcaversaccio

gakonst:
let's send it

PR #765, opened 2023-01-09
Eip-712 section should cover byte/string encoding

robriks:
Dynamic values of the EIP-712 struct hash (ie bytes and string types) must be encoded as a keccak256 hash of their contents. However since this tutorial codebase doesn't involve any bytes/string dynamic values, that aspect of 712 is not covered at all. It should be mentioned since it can confuse devs like me writing eip712 meta-tx contracts.

I wrote a brief informative blurb explaining why it's not shown here, why it's relevant regardless, and then linking to the exact line of the EIP spec gh repo where this is discussed.

### DIFF
--- a/src/tutorials/testing-eip712.md
+++ b/src/tutorials/testing-eip712.md
@@ -131,6 +131,10 @@ contract SigUtils {
 }
 ```
 
+### Handling Dynamic Values
+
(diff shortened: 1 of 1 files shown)


mattsse (approved):
useful

PR #775, opened 2023-01-16
`forge doc` documentation

### LINKED ISSUE #770: Document `forge doc`
https://github.com/foundry-rs/foundry/pull/2701

rkrasiuk:
closes #770


src/reference/config/doc-generator.md
```diff
@@ -0,0 +1,34 @@
+## Documentation Generator
+
+Configuration related to the behavior of the Forge documentation generator. These keys are set in `[doc]` section.
+
+##### `out`
+
+- Type: string
+- Default: `docs`
+- Environment: `FOUNDRY_DOC_OUT` or `DAPP_DOC_OUT`
```
mds1, line 9: Is it worth mentioning the `DAPP_` version? Since that's legacy from dapptools for other env vars, maybe we start deprecating it by just not mentioning it in the docs for new env vars (or removing support upstream, but that may be more work? not sure offhand how it's implemented)

rkrasiuk: sounds like a good idea, will remove

PR #825, opened 2023-02-20
feat(reference): add model checker config

wiasliaw:
reference for model checker config

relative PRs
- https://github.com/gakonst/ethers-rs/pull/2147
- https://github.com/gakonst/ethers-rs/pull/2123


wiasliaw:
It seems like the code blcok without labeling language or file will be treated as rust, which cause CI error.

mattsse (approved):

PR #846, opened 2023-03-10
Fix default depth in invariant runs from 256 to 15

0xLightt:
Default depth can be double checked here:
https://github.com/foundry-rs/foundry/blob/master/config/src/invariant.rs#L27

Edit:
Commits to foundry-rs changed the line that initializes the default depth. Prev line 30, now it is on line 27.
Default depth is still 15 though.


0xLightt:
kindly bumping this

PR #851, opened 2023-03-16
fix: add missing `--block` option to `cast age` examples

0x-r4bbit:
The examples of the `cast age` reference show non working commands due to the missing `--block` option. This commit fixes them.

### DIFF
--- a/src/reference/cast/cast-age.md
+++ b/src/reference/cast/cast-age.md
@@ -32,12 +32,13 @@ Get the timestamp of a block.
 
 1. Get the timestamp of the latest block:
     ```sh
-    cast age latest
+    cast age # or
+    cast age --block latest
     ```
 
 2. Get the timestamp of the genesis block:
     ```sh
-    cast age 1
+    cast age --block 1
     ```
 
 ### SEE ALSO


src/reference/cast/cast-age.md
```diff
@@ -32,12 +32,13 @@ Get the timestamp of a block.
...
     ```sh
-    cast age latest
+    cast age # or
+    cast age --block latest
     ```
 
 2. Get the timestamp of the genesis block:
     ```sh
-    cast age 1
+    cast age --block 1
```
mattsse, line 41: imo the --block argument is redundant in this context and I consider this a cast bug, which should accept unnamed block number input

mattsse: mind opening an issue on foundry instead?

0x-r4bbit: > imo the --block argument is redundant in this context and I consider this a cast bug, which should accept unnamed block number input

This was my first thought as well, but then, fixing the docs was a lower hanging fruit.

> mind opening an issue on foundry instead?

Will do!

mattsse:
closing this since `cast age` no longer requires --block argument

PR #864, opened 2023-04-08
fix: fix docs

wiasliaw:


src/reference/config/overview.md
```diff
@@ -16,7 +16,7 @@ You can select the profile to use by setting the `FOUNDRY_PROFILE` environment v
 
 ### Global configuration
 
-You can create a `foundry.toml` file in your home folder to configure Foundry globally.
+You can create a `.foundry.toml` file in `~/.foundry` folder to configure Foundry globally.
```
mattsse, line 19:
```suggestion
You can create a `foundry.toml` file in `~/.foundry` folder to configure Foundry globally.
```

mattsse (changes requested):

src/reference/config/overview.md
mattsse: ref 
https://github.com/foundry-rs/foundry/blob/c49e5e1b1df95214059a9c92a211a8ae81bc21dd/config/src/lib.rs#L1233-L1236

mattsse (approved):
ty

src/reference/config/overview.md
wiasliaw: Resolved.

PR #880, opened 2023-04-28
docs: ux fuzz invariant

0xMelkor:
Added documentation for this new foundry feature [#4744](https://github.com/foundry-rs/foundry/pull/4744).

### DIFF
--- a/src/reference/config/inline-test-config.md
+++ b/src/reference/config/inline-test-config.md
@@ -15,7 +15,7 @@ contract MyTest is Test {
 }
 ```
 
-What we are asking here is to run our fuzzer `100` and `500` times for the `default` and `ci` profiles respectively. The interesting fact is that this would override any fuzz `runs` setup existing at a broader context. Such a context would still provide configs for parameters not explicitely reported in in-line statements, acting then as a fallback.
+What we are asking here is to run our fuzzer `100` and `500` times for the `default` and `ci` profiles respectively. The interesting fact is that this would override any fuzz `runs` setup existing at a global level. All other configs would be inherited from the global context, making this acting as a fallback for all possible configurations.
 
 
 ### In-line fuzz configs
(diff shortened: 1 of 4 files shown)


src/reference/config/inline-test-config.md
```diff
@@ -0,0 +1,64 @@
...
+Foundry users are enabled to specify overall test configurations, using a combination of ENV variables and config statements in the `foundry.toml`. Checkout the [📚 Testing reference](./testing) for a detailed description.
+
+Despite this may work in the general case, some tests may need finer control over their configuration. For such reason Forge provides a way to specify per-test configs for invariant and fuzz testing scenarios. 
+
+Users can in-line test config statements directly in Solidity comments. This would affect the behavior of the `forge test` command for a specific test instance, as illustrated in the example below.
+
+```solidity
+contract MyTest is Test {
+  /// forge-config: default.fuzz.runs = 100
+  /// forge-config: ci.fuzz.runs = 500
```
mds1, line 11: thanks! just one question—we show triple slashes for all the comments here, but I think double slashes is supported too? If so let's show those in the docs, since triple usually means natspec comments, and we don't need these to be natspec

0xMelkor: Hi @mds1,
only natspec comments are supported by the current implementation. 
The parser in place reads a`ProjectCompileOutput` instance, that doesn't track double slashed comments.

If you need to support them, I can investigate it further. Just let me know.

mds1: ahh got it, so is there a lower bound on solidity versions this would be supported for? it doesn't use `/// @custom` tags so maybe not

0xMelkor: Since we are not using `/// @custom` we are safe.

Anyway block comments of this form are supported.
```solidity
    /**
     * forge-config: default.fuzz.runs = 1024
     * forge-config: default.fuzz.max-test-rejects = 500
     */
```

I need to extend docs

0xMelkor: Docs extended for block comments

mattsse (approved):
great!

ptal @Evalir

Evalir (approved):
lgtm! merging as this was already done on the foundry repo

PR #891, opened 2023-05-08
Update solmate-nft.md

portdeveloper:
Change test function names to obey rest of the foundry book

testSafeContractReceiver -> test_SafeContractReceiver


mattsse (approved):
make sense to make all this uniformly, but I'm not actually sure what the recommended style guide is

PR #895, opened 2023-05-12
feat(cheatcodes): document new expect call behavior

Evalir:
Follow up for https://github.com/foundry-rs/foundry/pull/4912


mattsse (approved):
lgtm

PR #905, opened 2023-05-19
Update dependencies.md

fetsorn:
Docs should say that remappings can be set in `foundry.toml`. Maybe they do somewhere else? Only found this in some issue that recommended `forge config --basic > foundry.toml`.


mattsse (approved):
lgtm

PR #907, opened 2023-05-21
Add Precompile Documentation and Registry

jtriley-eth:
Adds the precompile registry to document addresses that are either reserved or exhibit potentially unexpected behavior for testing purposes.

Additionally, an entry for `assumePayable` was added.

Previously, these cheats and the precompile addresses appeared to be undocumented.


mattsse (approved):
nice tysm

mattsse:
@all-contributors add  @jtriley-eth for docs

PR #911, opened 2023-05-27
Update load.md

crystalbit:
Looks like there should be `vm.load`, not `cheats.load`

Also I don't know why there is a change in line 28, but looks like it should be there


mattsse (approved):

mattsse:
https://github.com/all-contributors add @crystalbit for docs

PR #918, opened 2023-06-12
fix(docs): add missing flag to example

agostbiro:
Since https://github.com/foundry-rs/foundry/pull/4874 the `--from` flag only works in combination with the `--unlocked` flag.  This commit adds the missing `--unlocked` flag to an example that is using `--from`.


Evalir (approved):
lgtm! this particular example needed it.

A note though: you can use `--from` without `--unlocked`, but using `--unlocked` _requires_ `--from`.

PR #922, opened 2023-06-16
feat(`v1`): Migration guide

Evalir:
This PR contains the migration guide for V1. It contains:
- A guide on how to upgrade foundry locally to V1
- A list of deprecated features
- A list of non-breaking changes that are important to know
- A list of breaking changes, and how to tackle them
- Recommended tips for managing the migration.

I expect to flesh out more content on this over the next few days to make it more thorough, but I believe this draft covers enough for us to start discussing.

cc @mds1 for feedback!


src/getting-started/migrating-to-v1.md
```diff
@@ -0,0 +1,328 @@
...
+    /// CORRECT BEHAVIOR: mock.add will revert due to arithmetic errors,
+    /// and it will be successfully detected by the `expectRevert` cheatcode.
+    function testRevertOnMathLibWithMock() public {
+        vm.expectRevert();
+        mock.add(2 ** 128 - 1, 2 ** 255 - 1);
+    }
+}
+```
+
+### Recommended approach for tackling codebase migrations
```
Evalir, line 323: I think adding bite-sized pieces of before/after test cases would be cool

Evalir (commented):
a small self-review

note that i haven't done a grammar check on this yet. will do so tomorrow

src/migrating-to-v1/breaking-changes.md
```diff
@@ -0,0 +1,292 @@
...
+
+When it comes to breaking changes, there has been a slight rework for the assertion (`expect*`) cheatcodes that removes previously allowed unintended or confusing behavior.
+
+The biggest change is that these cheatcodes will now work only for the next call, not at the "test" level anymore. The philosophy behind this is that by making these cheatcodes more strict through this requirement, tests will be more accurate as the cheatcodes will now only succeed under more strict conditions.
+
+#### `expectEmit`
+
+`expectEmit` previously allowed you to assert that a log was emitted during the execution of the test. With V1, the following changes have been made:
+
+- it now only works for the _next call_. This means if you used to declare all your expected emits at the top of the test, you may now have to move them to just before the next call you perform. Cheatcode calls are ignored. As long as the events are emitted during the next call's context, they will be matched.
```
mds1, line 11: two small changes, first one is the same in both suggestions, second one differs a bit
```suggestion
- It now only works for the _next call_. This means if you used to declare all your expected emits at the top of the test, you may now have to move them to just before the next call you perform. Cheatcode calls are ignored. As long as the events are emitted within the next call's context, they will be matched.
```
```suggestion
- It now only works for the _next call_. This means if you used to declare all your expected emits at the top of the test, you may now have to move them to just before the next call you perform. Cheatcode calls are ignored. As long as the events are emitted during the next call's execution, they will be matched.
```

src/migrating-to-v1/breaking-changes.md
```diff
@@ -0,0 +1,292 @@
...
+To illustrate, see the following examples:
+
+```solidity
+
+contract Emitter {
+    event A(uint256 indexed a);
+    event B(uint256 indexed b);
+    event C(uint256 indexed c);
+    event D(uint256 indexed d);
+    event E(uint256 indexed e);
```
mds1, line 24: Nit, but maybe make the events a bit different to ensure no one gets confused by thinking this only applies in cases like this where everything is indexed or something
```suggestion
    event A(uint256 indexed a);
    event B(uint256 b);
    event C(address indexed c1, uin256 c2);
    event D(bytes32 d1, uint256 d2, address indexed d3);
    event E(uint256 indexed e);
```

src/migrating-to-v1/breaking-changes.md
```diff
@@ -0,0 +1,292 @@
...
+    function setUp() public {
+        emitter = new Emitter();
+    }
+
+    /// CORRECT BEHAVIOR: Declare all your expectEmits just before the next call,
+    /// on the test.
+    /// emit() emits [A, B, C, D, E], and we're expecting [A, B, C, D, E].
+    /// this passes.
+    function testExpectEmit() public {
+        cheats.expectEmit(true, false, false, true);
```
mds1, line 54: Should we `vm` here instead? I think the book may be inconsistent and use both, but `vm` is more common IME
```suggestion
        vm.expectEmit(true, false, false, true);
```

src/migrating-to-v1/breaking-changes.md
```diff
@@ -0,0 +1,292 @@
...
+    event E(uint256 indexed e);
+
+    function setUp() public {
+        emitter = new Emitter();
+    }
+
+    /// CORRECT BEHAVIOR: Declare all your expectEmits just before the next call,
+    /// on the test.
+    /// emit() emits [A, B, C, D, E], and we're expecting [A, B, C, D, E].
+    /// this passes.
```
mds1, line 52: Same for other "this passes" comments, just capitalizing since it's a new sentence. Let's also use `//` since there are regular non-natspec comments
```suggestion
    // This passes.
```

src/migrating-to-v1/breaking-changes.md
```diff
@@ -0,0 +1,292 @@
...
+    function testExpectEmit() public {
+        cheats.expectEmit(true, false, false, true);
+        emit A(1);
+        cheats.expectEmit(true, false, false, true);
+        emit B(2);
+        cheats.expectEmit(true, false, false, true);
+        emit C(3);
+        cheats.expectEmit(true, false, false, true);
+        emit D(4);
+        cheats.expectEmit(true, false, false, true);
```
mds1, line 62: We should also mix things up with the `expectEmit` args a bit, and show the `vm.expectEmit()` syntax too

src/migrating-to-v1/breaking-changes.md
```diff
@@ -0,0 +1,292 @@
...
+        emit B(2);
+        cheats.expectEmit(true, false, false, true);
+        emit E(5);
+        emitter.emit();
+    }
+
+    /// CORRECT BEHAVIOR: Declare all your expectEmits in an internal function.
+    /// Calling a contract function internally is a JUMPI, not a call,
```
mds1, line 95: This is probably just a `JUMP` not a `JUMPI` since there's no conditional

src/migrating-to-v1/breaking-changes.md
```diff
@@ -0,0 +1,292 @@
...
+    /// therefore it's a valid pattern, useful if you have a lot of events to expect.
+    /// emit() emits [A, B, C, D, E], and we're expecting [B, D, E].
+    /// this passes.
+    function testExpectEmitWithInternalFunction() public {
+        declareExpectedEvents();
+        emitter.emit();
+    }
+
+    function declareExpectedEvents() internal {
```
mds1, line 104: We should also have an incorrect example of e.g. a library function / free function where the event is emitted within the test contract and therefore can't be caught

src/migrating-to-v1/breaking-changes.md
```diff
@@ -0,0 +1,292 @@
+## Breaking changes
+
+When it comes to breaking changes, there has been a slight rework for the assertion (`expect*`) cheatcodes that removes previously allowed unintended or confusing behavior.
+
+The biggest change is that these cheatcodes will now work only for the next call, not at the "test" level anymore. The philosophy behind this is that by making these cheatcodes more strict through this requirement, tests will be more accurate as the cheatcodes will now only succeed under more strict conditions.
+
+#### `expectEmit`
```
mds1, line 7: Adding a table of contents so we can jump to different sections on this page would be nice

src/migrating-to-v1/breaking-changes.md
```diff
@@ -0,0 +1,292 @@
...
+
+```solidity
+contract Protocol {
+    function doSomething() external {
+        // do something...
+    }
+}
+
+contract Caller {
+    Protocol public protocol;
```
mds1, line 134: Seems we never initialize this variable anywhere

mds1 (commented):
Didn't finish reviewing yet but here's the first batch of comments. Really great overall so far though, will finish giving this a read tomorrow

src/migrating-to-v1/breaking-changes.md
mds1: Ah it seems you have a section for this later, though it's only for expectCall/Revert, not Emit. Maybe include expectEmit in that section too and link to it in each of the prior expectCall/Revert/Emit sections?

src/migrating-to-v1/deprecated-features.md
```diff
@@ -0,0 +1,67 @@
...
+    // Assert your invariants...
+}
+
+/// New
+function statefulFuzzTestEq() public  {
+    // Assert your invariants
+}
+```
+
+### Deprecating testFail
```
mds1, line 21:
```suggestion
### `testFail`
```

src/migrating-to-v1/deprecated-features.md
```diff
@@ -0,0 +1,67 @@
...
+
+Using `testFail` to write failing tests is now discouraged. While it's usage is extremely common, we've decided to deprecate it because it can introduce footguns in a test.
+
+The now-recommended pattern is to refactor the actions that will make your test revert and subsequently fail in an utility function, and use `expectRevert` to ensure that this function call failed. This way, the test failure is expected explicitly, instead of introducing the possibility that the test fails in an unintended way silently.
```
mds1, line 25: Similarly, I think this has always been recommended. And we can simplify this a bit since the "refactor" part may not be necessary and has already been covered in the expectRevert section
```suggestion
A better pattern is to use `expectRevert` to ensure that a function call reverted in a specific way. This way, the test failure is expected explicitly, removing the possibility that the test fails in an unintended, undetectable way.
```

src/migrating-to-v1/deprecated-features.md
```diff
@@ -0,0 +1,67 @@
...
+
+    /// Deprecated way.
+    function testFailReverter() public {
+        mock.revert_();
+    }
+
+    /// New way, without using testFail.
+    /// The call to revert_ has been refactored and is now expected to fail
+    /// with expectRevert()
+    function testReverterReverts() public {
```
mds1, line 52: We should also start making sure the test names in these code snippets match what's in the best practices page, for consistency
```suggestion
    function test_RevertIf_AlwaysReverts() public {
```

src/migrating-to-v1/deprecated-features.md
```diff
@@ -0,0 +1,67 @@
...
+        vm.expectRevert();
```
mds1, line 53: let's avoid using the empty args version in the examples, since (like `testFail`) this is weaker and should only be used when necessary

src/migrating-to-v1/deprecated-features.md
```diff
@@ -0,0 +1,67 @@
...
+    function exposed_callReverter() public {
+        mock.revert_();
+    }
+
+}
+```
+
+### Removing cheatcodes support on some precompiles
+
+It is now impossible to use the following cheatcodes on precompiles: `etch`, `load`, `store`, `deal`.
```
mds1, line 67: let's add rationale as to why

src/migrating-to-v1/migrating-to-v1.md
```diff
@@ -0,0 +1,15 @@
+## Migrating to Foundry V1
+
+Foundry V1 is the blazing-fast smart contract testing framework you already know, but finally marked as stable and respecting semver. It's the culmination of over a year of work and more than 200 contributors that chimed in to make it happen.
```
mds1, line 3:
```suggestion
Foundry V1 is the blazing-fast smart contract testing framework you already know, but finally marked as stable and respecting [semver](https://semver.org/). It's the culmination of over a year of work and more than 200 contributors that made it happen.
```

src/migrating-to-v1/migrating-to-v1.md
```diff
@@ -0,0 +1,15 @@
...
+
+With the V1 release, very few breaking changes have been included, meaning migration should be mostly painless. This guide will help you understand what these changes are and how to handle them properly.
+
+### What to expect
+
+V1 mainly brings more strictness to several cheatcodes, which will be the main thing to migrate early on. While we recommend migrating as soon as possible to V1, It's possible to upgrade your local installation and migrate the tests, while maintaining nightly usage on CI to avoid any failures that might disrupt your workflow.
```
mds1, line 9: link to docs on how to do what's suggested in that second paragraph? I'm a bit confused on that because wouldn't the nightlies just become the v1 version that have breaking changes? I'd have thought you need to pin CI to a pre-v1 nightly
```suggestion
V1 mainly increases the strictness of several cheatcodes, and accommodating this is the primary change required to migrate to V1. We recommend migrating to V1 as soon as possible, as the stricter cheats may reveal bugs or issues in your code and test suite. 

If migration is time consuming, you can consider upgrading your local foundry installation while migrating tests, but maintaining nightly usage on CI to avoid any failures that might disrupt your workflow.
```

src/migrating-to-v1/migrating-to-v1.md
```diff
@@ -0,0 +1,15 @@
...
+
+Aside from this, V1 is still the same Foundry you know.
+
+### I need support. Where do I go?
+
+The [dev Telegram](https://t.me/foundry_rs) is available for any concerns you may have that are not covered in this guide.
```
mds1, line 15: should we link to the support telegram here instead? or maybe both and specify when to use which

src/migrating-to-v1/non-breaking-changes.md
```diff
@@ -0,0 +1,6 @@
+## Non-breaking but important changes
+
+- Snapshot failures between fuzz runs are now persisted, so workarounds are not needed anymore.
```
mds1, line 3: what does this mean? maybe we should link to issues/PRs here for more context

src/migrating-to-v1/non-breaking-changes.md
```diff
@@ -0,0 +1,6 @@
...
+- Sensitive broadcast logs are now logged into the `.gitignore`d `cache` directory, instead of written to logs.
```
mds1, line 4:
```suggestion
- Sensitive broadcast logs are now logged into the `.gitignore`d `cache` directory, instead of being included in the broadcast files that are intended to be committed.
```

src/migrating-to-v1/upgrading-to-v1.md
```diff
@@ -0,0 +1,19 @@
...
+
+### Local Installation
+
+To upgrade, you'll need to reinstall `foundryup`. To do so, open your terminal and run the following command:
+
+```sh
+curl -L https://foundry.paradigm.xyz | bash
+```
+
+This will reinstall `foundryup`. You can then run `foundryup` on your terminal to download the latest `V1` version.
```
mds1, line 11: style nit: we use `V1` everywhere, do we prefer this or `v1`? I'm indifferent but usually use lowercase

src/migrating-to-v1/upgrading-to-v1.md
```diff
@@ -0,0 +1,19 @@
...
+
+After this, you're now up to date with the latest stable Foundry version.
+
+### Upgrading your CI
+
+When it comes to CI, you can add a `version` input field with `v1.0.0` or your desired stable release. This will pin the CI to that stable version.
```
mds1, line 17: does this support ways to specify a range of versions, e.g. `v1` to always use the latest `v1.x.y` release? we definitely should support something like that, which is common for github actions

Evalir: Ah, not really—but we could add this and release a new CI toolchain version.

src/migrating-to-v1/breaking-changes.md
```diff
@@ -0,0 +1,337 @@
...
+    function setUp() public {
+        emitter = new Emitter();
+    }
+
+    // CORRECT BEHAVIOR: Declare all your expectEmits just before the next call,
+    // on the test.
+    // emitEvent() emits [A, B, C, D, E], and we're expecting [A, B, C, D, E].
+    // This passes.
+    function testExpectEmit() public {
+        vm.expectEmit(true, false, false, true);
```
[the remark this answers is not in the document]
lucas-manuel: Not a breaking change though so all good if its not included

src/migrating-to-v1/breaking-changes.md
```diff
@@ -0,0 +1,337 @@
...
+}
+
+library EmitLib {
+    event Log();
+    function emit() internal {
+        emit Log();
+    }
+}
+
+// intermediate "mock" contract that calls the library function
```
lucas-manuel, line 282: Intermediate

src/migrating-to-v1/deprecated-features.md
```diff
@@ -0,0 +1,67 @@
+## Deprecated features
+
+With foundry v1, we've started deprecating features that are considered anti-patterns or might be misleading and confusing for developers. Deprecated means that while you can continue using these features, it is now a discouraged practice and you should plan on migrating away from using these.
+
+### Removing the Invariant keyword
```
lucas-manuel, line 5: `invariant` prefix?

src/migrating-to-v1/deprecated-features.md
```diff
@@ -0,0 +1,67 @@
...
+
+    /// Deprecated way.
+    function testFailReverter() public {
+        mock.revert_();
+    }
+
+    // Better way, without using testFail.
+    /// The call to revert_ has been refactored and is now expected to fail
+    /// with expectRevert()
+    function test_RevertIf_AlwaysReverts() public {
```
lucas-manuel, line 52: `test_revertIf`..

src/migrating-to-v1/deprecated-features.md
```diff
@@ -0,0 +1,67 @@
...
+        vm.expectRevert("This reverts);
```
lucas-manuel, line 53: "This reverts"

src/migrating-to-v1/deprecated-features.md
```diff
@@ -0,0 +1,67 @@
+## Deprecated features
+
+With foundry v1, we've started deprecating features that are considered anti-patterns or might be misleading and confusing for developers. Deprecated means that while you can continue using these features, it is now a discouraged practice and you should plan on migrating away from using these.
```
lucas-manuel, line 3: Foundry

src/migrating-to-v1/migrating-to-v1.md
```diff
@@ -0,0 +1,17 @@
...
+
+With the v1 release, very few breaking changes have been included, meaning migration should be mostly painless. This guide will help you understand what these changes are and how to handle them properly.
+
+### What to expect
+
+v1 mainly increases the strictness of several cheatcodes, and accommodating this is the primary change required to migrate to v1. We recommend migrating to v1 as soon as possible, as the stricter cheats may reveal bugs or issues in your code and test suite. 
+
+If migration is time consuming, you can consider upgrading your local foundry installation while migrating tests, but maintaining nightly usage on CI to avoid any failures that might disrupt your workflow.
+
+Aside from this, V1 is still the same Foundry you know.
```
lucas-manuel, line 13: v1

src/migrating-to-v1/non-breaking-changes.md
```diff
@@ -0,0 +1,6 @@
+## Non-breaking but important changes
+
+- Snapshot failures between fuzz runs [are now persisted](https://github.com/foundry-rs/foundry/pull/4974), so workarounds are not needed anymore when testing fuzz runs that use snapshots.
+- Sensitive broadcast logs are now logged into the `.gitignore`d `cache` directory, instead of being included in the broadcast files that are intended to be committed.
+- The base fee calculation on Anvil is now entirely skipped if the fee is set to 0.
+- `vm.startPrank` now overwrites the existing prank instead of just erroring. However, the already set prank must be used first.
```
lucas-manuel, line 6: `However, the already set prank must be used first.` this sentence wasn't immediately clear to me, what does this mean?

src/migrating-to-v1/upgrading-to-v1.md
lucas-manuel, line 11: We should standardize `V1`, V1 and v1. I'm indifferent as to which but should be the same across the docs.

lucas-manuel (changes requested):

src/migrating-to-v1/upgrading-to-v1.md
Evalir: Agreed. My pick is v1—but wdyt @gakonst ? needs to align with blog post.

src/migrating-to-v1/non-breaking-changes.md
Evalir: Hmm, I believe this might be an undocumented, somewhat recent change on startPrank—when this overwrite change was made, i believe the rule is now that if the prank has not been used, it will error. if it has been used, it will overwrite the existing prank. 

so essentially before overwriting the prank you need to use it first

src/migrating-to-v1/deprecated-features.md
Evalir: Done in https://github.com/foundry-rs/book/pull/922/commits/0622765622b80670ca2b838eff1b368440516b65

src/migrating-to-v1/migrating-to-v1.md
Evalir: Done in https://github.com/foundry-rs/book/pull/922/commits/0622765622b80670ca2b838eff1b368440516b65

src/migrating-to-v1/deprecated-features.md
Evalir: Done in https://github.com/foundry-rs/book/pull/922/commits/0622765622b80670ca2b838eff1b368440516b65

src/migrating-to-v1/deprecated-features.md
Evalir: Done in https://github.com/foundry-rs/book/pull/922/commits/0622765622b80670ca2b838eff1b368440516b65

src/migrating-to-v1/breaking-changes.md
Evalir: Done in https://github.com/foundry-rs/book/pull/922/commits/0622765622b80670ca2b838eff1b368440516b65

src/migrating-to-v1/deprecated-features.md
Evalir: Done in https://github.com/foundry-rs/book/pull/922/commits/0622765622b80670ca2b838eff1b368440516b65

great catch!

src/migrating-to-v1/breaking-changes.md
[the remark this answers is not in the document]
Evalir: it's a good suggestion! added the refresher here: https://github.com/foundry-rs/book/pull/922/commits/9f67b28d479691c83c76fd0832173d196dcfb109

src/migrating-to-v1/deprecated-features.md
mds1: The best practices page of the book uses `test_RevertIf_AlwaysReverts` style though

thedavidmeister:
please reconsider the changes to `expectRevert` being unable to test reverts behind a jump

or provide some functionality that allows me to specify the depth, if the intent is to be more strict i'm happy to be more explicit

Evalir:
Should this be closed? @zerosnacks

PR #925, opened 2023-06-18
fix: base-fee command docs

merklefruit:
Current behavior:

```
❯ cast basefee
error: unrecognized subcommand 'basefee'
```

Corrected the docs to spell the command as `base-fee` instead


mattsse (approved):
lgtm,

do you want to add an alias for this to cast directly? so `basefee` is also supported?

merklefruit:
@mattsse Sure!

PR #926, opened 2023-06-19
Update dependencies.md to use an actual contract in solmate-utils

mgranados:
Otherwise the forge build command fails


mattsse (approved):
ty

PR #927, opened 2023-06-20
Update learn-foundry.md

portdeveloper:
Update the links inside learn-foundry.md to open in a new tab. Used html since markdown does not support it.


mattsse (approved):
nice, ty!

PR #930, opened 2023-06-22
Task: Add default chain id to anvil reference

### LINKED ISSUE #929: Add chain id to anvil output
I'm going through Patrick Collin's Foundry course and am currently at the part where you link Anvil to MetaMask, I noticed the default chain id for Anvil isn't clearly stated in the documentation. Without the course I don't think I would have found the information. I propose adding this detail to the page below for better accessibility: 

https://book.getfoundry.sh/reference/anvil/ 

PS. If it is clearly stated and I'm blind ignore this 😛

jtardioli:
Closes #929 

image


Evalir (approved):
Thank you! ^^

PR #935, opened 2023-06-26
Update forge-script.md

nnnnicholas:
Explains how to verify contracts previously deployed via script -- something I didn't know until I read this tip from @apoorvlathey 

https://twitter.com/apoorvlathey/status/1673443599932076032

### DIFF
--- a/src/reference/forge/forge-script.md
+++ b/src/reference/forge/forge-script.md
@@ -109,4 +109,9 @@ Scripts can be used to apply state transitions on live contracts, or deploy and 
         --verify -vvvv --resume
     ```
 
+4. Verify contracts that were previously deployed with a script
+    ```sh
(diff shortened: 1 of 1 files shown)


src/reference/forge/forge-script.md
```diff
@@ -109,4 +109,9 @@ Scripts can be used to apply state transitions on live contracts, or deploy and
         --verify -vvvv --resume
     ```
 
+4. Verify contracts that were previously deployed with a script
```
Evalir, line 112:
```suggestion
4. Verify contracts that were just deployed with a script
```

Evalir (approved):
Thanks! This is actually more of a result of how scripts are processed and it's why it's possible. Made a suggestion to made sure that people know you have to do this _right after_ deploying with a script, but this should then be good.

PR #936, opened 2023-06-27
chore: update storage docs to include no slot

zobront:
Recently learned that `cast storage` can be used to return the full storage layout of a contract if verified on Etherscan.

This fact isn't included in the docs, so adding here.

### DIFF
--- a/src/reference/cast/cast-storage.md
+++ b/src/reference/cast/cast-storage.md
@@ -2,7 +2,7 @@
 
 ### NAME
 
-cast-storage - Get the raw value of a contract's storage slot or full storage layout.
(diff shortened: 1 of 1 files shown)


src/reference/cast/cast-storage.md
```diff
@@ -2,7 +2,7 @@
 
 ### NAME
 
-cast-storage - Get the raw value of a contract's storage slot.
+cast-storage - Get the raw value of a contract's storage slot or full storage layout.
```
Evalir, line 5:
```suggestion
cast-storage - Get the raw value of a contract's storage slot or its full storage layout.
```

Evalir (approved):
ty 👑 !

PR #947, opened 2023-07-06
feat: change SPDX identifier to MIT and upgrade pragma version

QEDK:
It should probably be under an open-source license since it's part of the documentation, also upgraded the version to the latest available.


mattsse (approved):
sgtm!

PR #951, opened 2023-07-09
docs: Update naming conventions for some `cast` commands

Sabnock01:
Previously, some `cast` ABI, Conversion, and Utility commands used a `--` prefix. This PR changes the documentation to reflect this as well as corrects the names of some others.

* `from-fix` -> `from-fixed-point`
* `to-fix` -> `to-fixed-point`
* `to-checksum-address` -> `to-check-sum-address` (The incorrect name makes more sense to me though actually)


mattsse (approved):
love it

tysm

PR #957, opened 2023-07-14
Named import {Test} 

tinotendajoe01:
Following the best practice for imports "We must always use named import syntax" Tests in the Projects needed to follow this standard of named imports i.e 

```
import { Test } from "forge-std/Test.sol";

```
instead of 
```
import   "forge-std/Test.sol";
```


tinotendajoe01:
Project Test files need to have an exclusive named import when using Test.sol

Evalir (changes requested):
Hey! appreciate this pr but it formats the whole codebase with your own settings which is not desirable. Ideally we'd only get the named imports.

PR #976, opened 2023-08-09
Add parseJsonKeys references

klkvr:
Adds docs for parseJsonKeys added in https://github.com/foundry-rs/foundry/pull/5252

PR #982, opened 2023-08-13
Fix example for `parseJsonKeys`

klkvr:
I have introduced incorrect example of cheatcode usage in #976, this PR fixes it


mattsse (approved):
nice, tysm!

PR #987, opened 2023-08-18
fix: add correct cargo command on installation guide

Evalir:
As we split the cli crate, we need to update this

https://github.com/foundry-rs/foundry/issues/5659#issuecomment-1684237995


mattsse (changes requested):
we also need to update the other insrtuctions

mattsse (approved):

PR #989, opened 2023-08-22
Cast subcommand syntax in README.md

robriks:
`cast --from-utf8` -> `cast from-utf8` 

Cast subcommand doesn't execute properly as shown due to outdated syntax


Evalir (approved):
thanks! will also update foundry to be able to use `--from-utf8`—should still be the case

PR #993, opened 2023-08-29
feat(`cheatcodes`): add docs for exists, isFile and isDir cheats

ruvaag:
Add docs for new filesystem cheatcodes `exists`, `isFile` and `isDir` from foundry-rs/foundry#5719
* Update `src/cheatcodes/fs.md` with cheatcode sigs
* Update `src/cheatcodes/README.md` with cheatcode sigs

### DIFF
--- a/src/cheatcodes/README.md
+++ b/src/cheatcodes/README.md
@@ -338,6 +338,15 @@ interface CheatCodes {
     // - The user lacks permissions to remove the file.
     // (path) => ()
     function removeFile(string calldata) external;
+    // Returns true if the given path points to an existing entity, else returns false
+    // (path) => (bool)
+    function exists(string calldata) external returns (bool);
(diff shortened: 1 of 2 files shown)


Evalir (changes requested):
Sweet, this so far lgtm. We also need a page for each cheatcode explaining what they do like https://github.com/foundry-rs/book/blob/master/src/cheatcodes/chain-id.md

ruvaag:
Got it, thanks. Looks like the filesystem cheatcodes are grouped together under [src/cheatcodes/fs.md](https://github.com/foundry-rs/book/blob/master/src/cheatcodes/fs.md) for now.

Would it be better to append an example block for `exists`, `isFile` and `isDir` on the same page to keep them grouped? If not, I can create separate pages for the three of these.

Evalir:
gotcha! yep, let's group them up on `fs.md`

Issue #179, opened 2022-04-07
`--match-path` is now a glob instead of regex


onbjerg:
See https://github.com/gakonst/foundry/pull/1229

PR #997, opened 2023-09-01
Update parse-json.md to reference `foundry.toml` permissions

0xGuybrush:
I was hitting into the same issue described in [issue #179](https://github.com/foundry-rs/forge-std/issues/179) and just found it through Google searching the error.

Just a minor suggestion to add the solution directly to the docs page


Evalir (approved):
ty!

PR #999, opened 2023-09-08
chore: add info about changing fuzz seed in CI

mds1:
cc @PaulRBerg since it links to sablier


Evalir (approved):
nice!

PaulRBerg:
Awesome, thanks for linking it to V2 Core!

PR #1003, opened 2023-09-11
Example usage of CREATE2 deployer proxy

saucepoint:
Providing an example of how to use the CREATE2 deployer proxy

### DIFF
--- a/src/reference/anvil/README.md
+++ b/src/reference/anvil/README.md
@@ -37,9 +37,35 @@ anvil --port <PORT>
 ```  
 
 #### Default CREATE2 Deployer
-Anvil, when used without forking, includes the [default CREATE2 deployer proxy](https://github.com/Arachnid/deterministic-deployment-proxy) at the address `0x4e59b44847b379578588920ca78fbf26c0b4956c`.
+Anvil, when used without forking, includes the [default CREATE2 deployer proxy](https://github.com/Arachnid/deterministic-deployment-proxy) at the address `0x4e59b44847b379578588920cA78FbF26c0B4956C`.
 
 This allows you to test CREATE2 deployments locally without forking.
+
+```solidity
+address CREATE2_DEPLOYER = address(0x4e59b44847b379578588920cA78FbF26c0B4956C);
+
+uint256 salt = 100;
+
+// Counter accepts uint256 in its constructor
+bytes memory creationCode = abi.encodePacked(type(Counter).creationCode, abi.encode(uint256(200)));
+
+// Precomputing the address of the contract
+address counterAddress = address(
+    uint160(uint256(keccak256(abi.encodePacked(
+        bytes1(0xff),
+        CREATE2_DEPLOYER,
+        salt,
+        keccak256(creationCode)
+    ))))
+);
+
+// deploying `Counter` with CREATE2 and salt
+(bool success, ) = CREATE2_DEPLOYER.call(
+    abi.encodePacked(salt, creationCode)
+);
+require(success, "Deployment failed");
+require(counterAddress.code.length > 0, "Contract not deployed");
+```
 
 #### Supported RPC Methods
 ##### Standard Methods


src/reference/anvil/README.md
```diff
@@ -37,10 +37,36 @@ anvil --port <PORT>
...
+    ))))
+);
+
+// deploying `Counter` with CREATE2 and salt
+(bool success, ) = CREATE2_DEPLOYER.call(
+    abi.encodePacked(salt, creationCode)
+);
+require(success, "Deployment failed");
+require(counterAddress.code.length > 0, "Contract not deployed");
+```
```
mds1, line 68: I don't understand what this code snippet is for? From solidity scripts and tests, you should never need to manually encode this way if you want to use the create2 deploy. `new Counter{salt: mySalt}()` is all you need

saucepoint: oh interesting, TIL!

Looks like `new Counter{salt: mySalt}()` works as expected in a script running against anvil, but doesnt seem to work in forge-test. The snippet I provided works in both

mds1: It should work in `forge test` as long as the deployer is there locally. If it's not, I'd suggest "deploying" it with `vm.etch("7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3")`, where that code is the deployed bytecode from https://etherscan.io/address/0x4e59b44847b379578588920ca78fbf26c0b4956c.

saucepoint: isnt the deployer already available in `forge test` if the verbose snippet works? either way tried etching the bytecode and still getting smacked around 🤕

script (works): https://github.com/saucepoint/v4-template/blob/create2-test/script/Counter.s.sol#L23-L36

test (fails): https://github.com/saucepoint/v4-template/blob/create2-test/test/Counter.t.sol#L36-L48

mds1: The test is failing from `HookAddressNotValid` though, which seems unrelated to the create2 deployer? (unless the address is being computed wrong and using the wrong deployer address)

saucepoint: was mostly confused why the same code in `forge script` and `forge test` had different outcomes. Turns out calculating salts / precomputed addresses have different behaviors in script and test

---

Assume:
```solidity
    /// @notice Precompute a contract address deployed via CREATE2
    function computeAddress(address deployer, uint256 salt, bytes memory creationCode) public pure returns (address) {
        return address(
            uint160(uint256(keccak256(abi.encodePacked(bytes1(0xff), deployer, salt, keccak256(creationCode)))))
        );
    }
 ```

in forge script you need to provide `deployer = 0x4e59b44847b379578588920cA78FbF26c0B4956C`

in forge test you need to provide `deployer = address(this)` (or the pranking address)

---

willing to update the docs with my findings if you think this is worthwhile information

Comment (mds1) [reply src/reference/anvil/README.md:68]: Ah, yea this is a known issue and we should update tests to use the same create2 deployer as scripts, ref https://github.com/foundry-rs/foundry/issues/5529. cc @Evalir

**Review by muntahafayiz — APPROVED** [review]

File: src/reference/anvil/README.md [hunk 2]
```diff
@@ -37,10 +37,36 @@ anvil --port <PORT>
...
+
 #### Supported RPC Methods
 ##### Standard Methods
 The standard methods are based on [this](https://eth.wiki/json-rpc/API) reference.
```
muntahafayiz, line 72: نسخة مكررة من #

src/reference/anvil/README.md
mds1: This is now resolved with the `always_use_create_2_factory` flag introduced in https://github.com/foundry-rs/foundry/pull/6656

onbjerg:
Closing this as stale, feel free to re-open it if progress is made on it:)

PR #1006, opened 2023-09-15
(WIP) feat(cheatcodes): add doc for cool

ditto-eth:
Ref https://github.com/foundry-rs/foundry/pull/5830
Ref https://github.com/foundry-rs/forge-std/pull/456

don't merge yet since still wip (old docs), just wanted to share for initial feedback since might change the api a bit!

### DIFF
--- a/src/cheatcodes/cool.md
+++ b/src/cheatcodes/cool.md
@@ -1,0 +1,67 @@
+## `cool`
+
+### Signature
+
+```solidity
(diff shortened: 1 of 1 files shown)


Evalir (changes requested):
Sweet! The docs look good bar any API change we might make. Let's also add the page to `SUMMARY.md` or else it won't show up.

PR #1004, opened 2023-09-13
cast wallet import / list

rplusq:
## Reason

After https://github.com/foundry-rs/foundry/pull/5551 was merged, there was a need to update the book.

Also, inspired by [this post](https://x.com/patrickalphac/status/1701627568082809009) where users were asking where to learn more.

I also added some recommendations on the Best Practices section, also inspired by the post + https://github.com/smartcontractkit/full-blockchain-solidity-course-js/discussions/5

## Content

- References for cast
- Updated options
- Added a new section to best practices

PR #1010, opened 2023-09-20
Update best practices around private keys

mds1:
In https://github.com/foundry-rs/book/pull/1004, a section on private key management was added to the best practices page. I think it's a bit too opinionated for the best practices page, so I updated it to be more balanced. 

In general, we should be careful about updating that page—it was originally a collaboration between a few teams to reach best practices we could all agree on, so we should maintain the same level of scrutiny any time that page is changed. Some things are objectively valuable and should always be followed—such as testing scripts—so we can state what to do there confidently and prescriptively. Other things are more subjective and everyone has their own approaches—such as key management—so we should be less prescriptive there and convey the nuances/tradeoffs of various options.

Also @Evalir can you fix the `../reference/common/wallet-options-hardware.md` link? It doesn't work but the path to that fike is correct. I'm assuming that page is imported and doesn't directly have it's own page, but I'm not sure where the right page is

cc @rplusq who opened #1004

### DIFF
--- a/src/tutorials/best-practices.md
+++ b/src/tutorials/best-practices.md
@@ -14,15 +14,8 @@ In general, it's recommended to handle as much as possible with [`forge fmt`](..
       - [Workaround Functions](#workaround-functions)
     - [Best practices](#best-practices-1)
     - [Taint Analysis](#taint-analysis)
-  - [Private Key Management](#private-key-management)
-    - [Safeguarding Your Private Keys](#safeguarding-your-private-keys)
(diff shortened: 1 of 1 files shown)


Evalir:
hey @mds1 — this link didn't work because it's not meant to be shown as a standalone page, but to be included where this option's available. I've changed the link to point to `forge script`'s reference.

Evalir (approved):
should be good to go!

mds1:
Thank you!

PR #1029, opened 2023-10-25
Document BTreeMap lexicographical order

robriks:
Added short blurbs on Rust's BTreeMap crate ordering system, specifying the caveat that Solidity structs must be declared alphabetically *except that uppercase and lowercase strings are treated differently where uppercase characters precede lowercase ones.

This should add clarity for developers using Foundry's nifty json parsing functionality :)


Evalir (approved):
thanks! this is correct—BTreeMap will produce a lexicographical order.

PR #1052, opened 2023-11-22
fix: cast proof slot should be padded to 32 bytes

excaliborr:
- cast proof seems to be expecting the `bytes32(uint256(0))` and not just the `uint256(0)` to get a specific storage slot proof

### DIFF
--- a/src/reference/cast/cast-proof.md
+++ b/src/reference/cast/cast-proof.md
@@ -49,7 +49,7 @@ The displayed output is a JSON object with the following keys:
 
 1. Get the proof for storage slot 0 on the WETH contract:
     ```sh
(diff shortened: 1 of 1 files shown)


mattsse (approved):
yep

mattsse:
actually, this isn't necessary https://github.com/foundry-rs/foundry/pull/6403

excaliborr:
> actually, this isn't necessary [foundry-rs/foundry#6403](https://github.com/foundry-rs/foundry/pull/6403)

The thing is the proofs that are generated from 0 and bytes32(0) are not the same, I found this when verifying them against an MPT where the proof from cast proof 0 reverts and cast proof bytes32(0) was a valid proof, thats why I was updating the documentation

PR #1068, opened 2023-12-13
docs: removed Files link in cheatcode ref/external

kquirapas:
Not sure if this was intended, but saw a misplaced `Files` reference link in [Cheatcodes Reference/External ](https://book.getfoundry.sh/cheatcodes/external)(See Screenshot Below)

oss

### DIFF
--- a/src/cheatcodes/external.md
+++ b/src/cheatcodes/external.md
@@ -16,4 +16,3 @@
 - [`envString`](./env-string.md)
 - [`envBytes`](./env-bytes.md)
 - [`parseJson`](./parse-json.md)
-- [Files](./fs.md)
--- a/src/cheatcodes/external.md
+++ b/src/cheatcodes/external.md
@@ -16,5 +16,4 @@
 - [`envString`](./env-string.md)
 - [`envBytes`](./env-bytes.md)
 - [`parseJson`](./parse-json.md)
-- [`parseJson`](./parse-json.md)
 - [`fs`](./fs.md)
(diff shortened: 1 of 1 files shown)


src/cheatcodes/external.md
```diff
@@ -16,4 +16,3 @@
 - [`envString`](./env-string.md)
 - [`envBytes`](./env-bytes.md)
 - [`parseJson`](./parse-json.md)
-- [Files](./fs.md)
```
Evalir, line 19: So, this is not misplaced, just named very differently to all the other files :) This should be called `fs`.

Evalir (changes requested):

src/cheatcodes/external.md
kquirapas: So it should be [`fs`] ?

Evalir: correct

kquirapas: Pushed new changes

src/cheatcodes/external.md
```diff
@@ -16,4 +16,5 @@
...
+- [`parseJson`](./parse-json.md)
```
Evalir, line 19:
```suggestion
```

Evalir (changes requested):

Evalir (approved):

kquirapas:
Thanks for reviewing and correcting @Evalir

PR #1075, opened 2023-12-24
Mention Foundry dev docs in Contributing

DaniPopes:


teddav:
thanks for letting me know @DaniPopes 
do you think i should update it? or is the Foundry doc good enough now?

DaniPopes:
@teddav The process is a lot simpler now, if you could link the dev docs that would be great

PR #1078, opened 2024-01-04
docs: add example to /reference/cast/cast-send

0xernesto:
Even though there is technically an example of how to add hex data to the `input` field of the transaction object [here](https://book.getfoundry.sh/cast/), it took me a long time to find this. This PR adds an example of how to do this directly on the `cast-send` page.


0xernesto:
@all-contributors add @0xernesto for docs

PR #1090, opened 2024-01-17
Fix: update AccountAccessKind

kopy-kat:
The enum `AccountAccessKind` is currently out of sync with the enum in foundry: https://github.com/foundry-rs/forge-std/blob/36c303b7ffdd842d06b1ec2744c9b9b5fb3083f3/src/Vm.sol#L26-L49


mattsse (approved):
nice,ty!

PR #1092, opened 2024-01-20
fix: allow empty command descriptions

DaniPopes:
https://github.com/foundry-rs/foundry/pull/6866

### DIFF
--- a/scripts/gen_output/help.py
+++ b/scripts/gen_output/help.py
@@ -191,7 +191,10 @@ def parse_description(s: str):
     idx = s.find("Usage:")
     if idx < 0:
         return "", s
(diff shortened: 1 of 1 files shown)


cdump:
Maybe it would be better if we enforced writing command descriptions?
All existing commands have descriptions, and I believe it's beneficial for documentation.

DaniPopes:
Sure, at the review level in Foundry or maybe with `#![warn(missing_docs)]`, not in an external repo.

cdump:
Until this check is implemented and fixed at the Foundry repo level, wouldn't it be better to identify and address the issue here, in another repo? It's not an ideal solution, but probably better than 'silently ignore.'

PR #1093, opened 2024-01-20
Update autogenerated docs

cdump:
Foundry build from commit d0a4452212f54cfe6b6ef03cf4f92c013e38af04

Also fixes gen_output scripts: `tree` could be not installed by default (on arch linux, for example) and `npm install` (or `bun install`) is required for current PaulRBerg/foundry-template master branch


mattsse (approved):
ty!

PR #1098, opened 2024-01-24
Update forking-mainnet-with-cast-anvil.md

prasincs:
Changes to a different random account since the example `$LUCKY_USER` now has 0 balance This also changes `cast send` to not take `--unlocked` param since it doesn't seem to be required anymore. I just ran through the steps again and updated the values


onbjerg (approved):
lgtm

PR #1112, opened 2024-02-07
Revert "Remove unused error message argument in example code (#1091)"

mds1:
This reverts commit dbefc9af30ebd288fbf1ef5dd4e568113196c8ad (from https://github.com/foundry-rs/book/pull/1091)

Discussed with @Evalir in discord and this should work as expected, cc @maurelian

### DIFF
--- a/src/cheatcodes/expect-revert.md
+++ b/src/cheatcodes/expect-revert.md
@@ -42,7 +42,7 @@ There are 3 signatures:
 >
 > ```solidity
 > function testLowLevelCallRevert() public {
->     vm.expectRevert();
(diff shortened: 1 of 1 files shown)


Evalir (approved):
lgtm—interested in @maurelian thoughts though? if you saw this behavior not work in tests, can we have a minimal repro?

mds1:
Chatted with him offline, it was a few weeks ago and no minimal repro currently—may have been a temporary regression, or may have been just a misunderstanding when testing things out

mattsse (approved):

PR #1117, opened 2024-02-14
[fix] Fix broken code blocks

WardenJakx:
- Fixes improper blocks from https://book.getfoundry.sh/forge/writing-tests

# Before
Pasted Graphic 9

# After
Pasted Graphic 10


DaniPopes (approved):
Thanks!

Issue #1122, opened 2024-02-16
Output Correction `cast rpc anvil_impersonateAccount $LUCKY_USER`


twynne20:
The output when running  `cast rpc anvil_impersonateAccount $LUCKY_USER` within [Forking Mainnet with Cast and Anvil
](https://github.com/foundry-rs/book/blob/master/src/tutorials/forking-mainnet-with-cast-anvil.md) is not reflective of the actual output. Albeit a minimal difference, may be confusing to new user. 

Suggesting a change from: 
```
# This calls Anvil and lets us impersonate our lucky user
$ cast rpc anvil_impersonateAccount $LUCKY_USER
$ cast send $DAI \
--from $LUCKY_USER \
  "transfer(address,uint256)(bool)" \
  $ALICE \
  300000000000000000000000 \
  --unlocked
blockHash               0xbf31c45f6935a0714bb4f709b5e3850ab0cc2f8bffe895fefb653d154e0aa062
blockNumber             15052891
...
```

to the following: 
```
# This calls Anvil and lets us impersonate our lucky user
$ cast rpc anvil_impersonateAccount $LUCKY_USER
null
$ cast send $DAI \
--from $LUCKY_USER \
  "transfer(address,uint256)(bool)" \
  $ALICE \
  300000000000000000000000 \
  --unlocked
blockHash               0xbf31c45f6935a0714bb4f709b5e3850ab0cc2f8bffe895fefb653d154e0aa062
blockNumber             15052891
...
```

Evalir:
See https://github.com/foundry-rs/book/pull/1123#issuecomment-1949488528

PR #1123, opened 2024-02-16
Update forking-mainnet-with-cast-anvil.md

twynne20:
Implementing change for Output Correction: `cast rpc anvil_impersonateAccount $LUCKY_USER` to properly reflect the `null` output 

Documented the issue here:
https://github.com/foundry-rs/book/issues/1122


Evalir:
Thanks, but I disagree—I think the $ clearly represents they're two separate commands. Having the output could just feel more confusing. Thanks for the contribution, but will close for now

PR #1126, opened 2024-02-23
Small fixes to the "Forking Mainnet" tutorial

lancelot-c:
The address used for the "LUCKY_USER" environment variable is different from the address retrieved from Etherscan although they should be the same. I kept the second one because it holds DAI whereas the first one don't.

Also, it makes more sense to name this user "unlucky" instead of "lucky" because we are basically stealing funds from him/her. It can be confusing for anyone reading the tutorial, as it was confusing for me at first.

### DIFF
--- a/src/tutorials/forking-mainnet-with-cast-anvil.md
+++ b/src/tutorials/forking-mainnet-with-cast-anvil.md
@@ -33,7 +33,7 @@ $ cast call $DAI \
 0
 ```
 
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
this should then also update other mentions of `lucky`

mattsse (approved):

Evalir (approved):

PR #1129, opened 2024-02-26
Nested Struct encoding example

### LINKED ISSUE #1095: feat(`docs`): Add more complex/nested encoding examples
Right now our encoding examples are pretty basic. We could do better by showing some complex examples, such as structs, nested structs, so on and so forth. This would be ideally useful for any deployments tools, `cast call`, and `cast send`.

kamuik16:
Closes #1095

### DIFF
--- a/src/misc/struct-encoding.md
+++ b/src/misc/struct-encoding.md
@@ -62,7 +62,7 @@ The ABI of the `f` function in this contract is:
 
 which reads: The function `f` takes 1 input of type `tuple` with two components of type `address` and `uint256`.
 
-**Complex/Nested Encoding:**
+**Nested Struct Encoding:**
 Here's a more complex example with nested structs:
 
 ```solidity
--- a/src/misc/struct-encoding.md
+++ b/src/misc/struct-encoding.md
@@ -61,3 +61,72 @@ The ABI of the `f` function in this contract is:
 ```
 
 which reads: The function `f` takes 1 input of type `tuple` with two components of type `address` and `uint256`.
+
+**Complex/Nested Encoding:**
+Here's a more complex example with nested structs:
+
+```solidity
+pragma solidity 0.8.21;
+
+contract Test {
+    struct nestedStruct {
+        address addr;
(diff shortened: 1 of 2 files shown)


kamuik16:
Ignore the cast balance commit. It already got merged in another PR https://github.com/foundry-rs/book/pull/1128.

src/misc/struct-encoding.md
```diff
@@ -61,3 +61,72 @@ The ABI of the `f` function in this contract is:
...
+            ]
+        }
+    ],
+    "name": "f",
+    "outputs": [],
+    "stateMutability": "pure",
+    "type": "function"
+}
+```
+which reads: The function `f` takes 1 input of type tuple with three components: a string, a uint256, and another tuple representing the nested struct with components addr of type address and amount of type uint256.
```
onbjerg, line 132: can we add an example of what this looks like when invoked? otherwise lgtm

onbjerg: like how you would pass it to e.g. constructor args for `forge create`, or for abi encoding with `cast abi-encode`

kamuikatsurgi: > like how you would pass it to e.g. constructor args for `forge create`, or for abi encoding with `cast abi-encode`

working on it.

kamuikatsurgi: > like how you would pass it to e.g. constructor args for `forge create`, or for abi encoding with `cast abi-encode`

added examples for both `forge create` and `cast abi-encode`.

kamuik16: > > like how you would pass it to e.g. constructor args for `forge create`, or for abi encoding with `cast abi-encode`
> 
> added examples for both `forge create` and `cast abi-encode`.

accidentally replied from another acc.

kamuik16: > like how you would pass it to e.g. constructor args for `forge create`, or for abi encoding with `cast abi-encode`

added examples for both `forge create` and cast `abi-encode`.

PR #1132, opened 2024-03-06
fix default evm version (london=>paris)

MathisGD:
It seems to me that the default evm_version is actually paris, and not london.

https://github.com/foundry-rs/foundry/blob/ce22450e4d625d12ff88fae347a68e3d3d9d2b61/crates/config/src/lib.rs#L1856


mattsse (approved):
thanks

DaniPopes (approved):

PR #1135, opened 2024-03-07
Add weekly automated command output update workflow 

zerosnacks:
This PR adds a weekly automated update workflow to more regularly update the command output in the documentation as command flags are added quite regularly in Foundry. You can also trigger this workflow manually.

The automated PR looks like this: https://github.com/zerosnacks/book/pull/3


mattsse (approved):
nice!
this is very useful

PR #1172, opened 2024-04-17
Repositioned info box at page 'Deterministic deployment using CREATE2'  from bottom to top.

7Cedars:
A small change: re-positioned the info box ('This guide is intended to help understand `CREATE2`. [...]) from the bottom of the page to the top. It is now positioned just after the introductory paragraph. 

Reason: the reader needs this information before reading through the page, not after.  

Hope that makes sense & thanks to for all for the work on Foundry book. It's an amazing resource.


onbjerg (commented):
this makes sense

onbjerg (approved):

PR #1173, opened 2024-04-19
mention solidity scripting as an alternative approach to deploying/verifying

stitetsati:
New Foundry users coming to the documentation wanting to find tutorials on deploying and verifying will always go to the "Deploying and Verifying" section of the doc because it's one of the top sections and sits at a more conspicuous location , but it only mentions the CLI approach. This would mislead the new devs to believe that there's no other way to deploy/verify multiple smart contracts using foundry. By mentioning the Solidity Scripting approach, the documentation will better inform the devs with the multiple approaches available to them

(I personally have been believing that Foundry's deployment CLI is the only way to deploy smart contracts with foundry and have been using a hybrid of hardhat + foundry as my setup...because its just impractical to use the CLI for deploying mid-large size projects with multiple contracts. It was not until very recently did I find out about the solidity scripting approach; now I only use foundry)

### DIFF
--- a/src/forge/deploying.md
+++ b/src/forge/deploying.md
@@ -4,7 +4,7 @@ Forge can deploy smart contracts to a given network with the [`forge create`](..
 
 Forge CLI can deploy only one contract at a time.
 
-Another way to deploy and verify smart contracts is to use Forge's [Solidity scripting](../tutorials/solidity-scripting.md#deploying-our-contract)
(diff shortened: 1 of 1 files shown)


mattsse (approved):

stitetsati:
@all-contributors please add @stitetsati for contributing doc

PR #1179, opened 2024-04-20
Update working-on-an-existing-project.md

shengwubin:
Add bun install to avoid compilation errors. I got some errors when building the project:

```shell
$ forge build
[⠃] Compiling...2024-04-20T11:16:26.701267Z ERROR foundry_compilers::artifacts: error="/path/to/my-project/node_modules/forge-std/src/Script.sol": No such file or directory (os error 2)
[⠊] Compiling...
Error: 
failed to resolve file: "/path/to/my-project/node_modules/forge-std/src/Script.sol": No such file or directory (os error 2); check configured remappings
	--> /path/to/my-project/script/Base.s.sol
	forge-std/src/Script.sol

```

ref: https://github.com/PaulRBerg/foundry-template


mattsse (approved):
makes sense for this example

PR #1216, opened 2024-06-07
add a note on Vyper support limitations

klkvr:


src/config/vyper.md
```diff
@@ -112,4 +112,9 @@ def run(counter: address):
...
 ```bash
 forge script script/Increment.s.vy  --sig 'run' '<counter address>' --rpc-url $RPC_URL --broadcast  --private-key $PRIVATE_KEY
-```
\ No newline at end of file
+```
+
+### 5. Limitations
+
+- While you can write and run tests and scripts in Vyper, there is no `new` keyword in Vyper allowing you to deploy contracts. This will be addressed in the future with new cheatcodes.
+- `forge coverage` currently does not support Vyper contracts.
```
DaniPopes, line 120: maybe on overloaded methods too?

klkvr: added

DaniPopes (approved):

mattsse (approved):

PR #1217, opened 2024-06-08
Fix confusing error in remappings docs.

aijams:
When trying to write a `remappings.txt` file for a project I was working on, I encountered a lot of `File not found` errors when running `foundry build`. Following the example in the docs that I intend to edit led me to believe that each remapping in my `remappings.txt` file should not begin with a '@' symbol. Only after trying to build my contracts with that symbol at the start of my remapping lines did `forge` manage to follow the remappings and find the files I intended it to find.

I was using `forge 0.2.0` on WSL (Windows subsystem for Linux).

I would like to spare future developers this headache when they start learning Foundry.


mattsse (approved):
ty, this was indeed wrong

PR #1220, opened 2024-06-14
chore: add valid error codes

ayenesimo1i:
## Description
added more valid error codes to solidity compiler config reference

error codes fetched from [here](https://github.com/foundry-rs/foundry/blob/master/crates/config/src/error.rs#L151-L197)


DaniPopes (approved):
Thanks!

mattsse (approved):

PR #1225, opened 2024-06-25
Use release profile for installing from sources

grandizzy:
Update profile with `release` since `local` was removed

### DIFF
--- a/src/getting-started/installation.md
+++ b/src/getting-started/installation.md
@@ -57,7 +57,7 @@ foundryup --path path/to/foundry
 Or, by using a single Cargo command:
 
 ```sh
-cargo install --git https://github.com/foundry-rs/foundry --profile local --locked forge cast chisel anvil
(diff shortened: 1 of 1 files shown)


mattsse (approved):

src/getting-started/installation.md
```diff
@@ -67,13 +67,13 @@ Or, by manually building from a local copy of the [Foundry repository](https://g
...
+cargo install --path ./crates/forge --profile release --force --locked
 # install Cast
-cargo install --path ./crates/cast --profile local --force --locked
+cargo install --path ./crates/cast --profile release --force --locked
 # install Anvil
-cargo install --path ./crates/anvil --profile local --force --locked
+cargo install --path ./crates/anvil --profile release --force --locked
 # install Chisel
-cargo install --path ./crates/chisel --profile local --force --locked
+cargo install --path ./crates/chisel --profile release --force --locked
```
DaniPopes, line 76: specifying the profile shouldn't be necessary as it's the default

PR #1227, opened 2024-06-27
fix: move precompile registry into `assumeNotPrecompile`

DaniPopes:
Its naming and the description are incredibly misleading since these are NOT supported by Foundry at the execution level, but only at the `assumeNotPrecompile` functions.


mattsse (approved):

mattsse (commented):
some link is incorrect

PR #1230, opened 2024-07-02
comma-separated -> semicolon-separated

dantastisk:
This does not work with a comma-separated list, only semicolon-separated, at least on macos using

```sh
$ forge -V
forge 0.2.0 (9058812 2024-07-01T00:25:00.180987000Z)
```


dantastisk:
nvm this, the semicolon obviously just ended the command 😅😅. Will open an issue instead

PR #1248, opened 2024-08-01
docs(`forge`): document non-obvious traces

leovct:
## Description

Document non-obvious traces that user might face regularly.

Related to https://github.com/foundry-rs/foundry/issues/7463


mattsse (approved):
ty!

PR #1253, opened 2024-08-05
Missing dependencies section in Soldeer

### LINKED ISSUE #1252: Missing dependencies section in Soldeer
> Note: I've only read the docs and noticed a discrepancy. I have not verified it. If it's invalid then feel free to close.

In `soldeer.md` there is a heading `Add a Dependency Stored in the Central Repository` which contains a `toml` block.
In the block it states that the following line must be added

```bash
libs = ["dependencies"] # <= This is important to be added
```

Scrolling down a little to `Updating Dependencies` there is a similar `toml` block and comment. The difference is that the line is missing but the comment is pointing at a different line.

```bash
gas_reports = ["*"] # <= This is important to be added
```

It appears that this may have been a mistake and that line was removed while the comment remained.

Solution: Add the line back in and move the comment.

Braqzen:
Added the following line to the `toml` block for `Updating Dependencies`

```bash
libs = ["dependencies"] # <= This is important to be added
```

Closes #1252

### DIFF
--- a/src/projects/soldeer.md
+++ b/src/projects/soldeer.md
@@ -87,7 +87,8 @@ For example, having this Foundry config file in a git repository, one can pull t
 auto_detect_solc = false 
 bytecode_hash = "none" 
 fuzz = { runs = 1_000 } 
-gas_reports = ["*"] # <= This is important to be added
+libs = ["dependencies"] # <= This is important to be added
+gas_reports = ["*"]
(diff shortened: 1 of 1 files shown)


mattsse (approved):

Braqzen:
@all-contributors add @Braqzen for doc

PR #1270, opened 2024-08-31
Mentioned deals' erc20 override in stdcheats in cheat reference

krzkaczor:
I think it's a little bit too buried in deal's docs that it actually possible to set erc20 balance.


onbjerg:
does not belong here, esp since we can't link to it in this section

krzkaczor:
I think that discoverability of the fact that foundry can set erc20 balances (which is much more harder than setting simply eth balances from technical pov) is very bad. It's only documented on details of `deal` cheatcode.

I know it doesn't fully belong there but as I said, it's totally not obvious that it's possible. Please reconsider.

Issue #1149, opened 2024-03-21
Foundry book contains invalid command for witnessing lib structure after dependency installation


0xumarkhatab:
# Main issue
[Chapter 5 of foundry book](https://book.getfoundry.sh/projects/dependencies#adding-a-dependency) contains an invalid command for latest version of foundry which when ran , gives following error message

![image](https://github.com/foundry-rs/book/assets/71306738/67fe2b6c-49dd-45a2-8e81-a9377fdefa22)

## To Reproduce
- open Powershell, in a foundry project.
- Install some dependency like solmate ( ` forge install transmissions11/solmate` )
- run `tree lib -L 1` at the root level

# My system specs
## Forge Version
 forge 0.2.0 (9fe9a3fc 2024-02-22T13:23:11.369603400Z)
## Terminal
Powershell
## OS 
Windows 10
## Rustc Version
rustc 1.76.0 (07dca489a 2024-02-04)
## Cargo Version
cargo 1.76.0 (c84b36747 2024-01-18)

zerosnacks:
Unable to replicate on Linux / MacOS, this is likely because the Windows `tree` command likely uses different parameters and doesn't recognize `-L`.

Feel free to open a PR for this documenting the behaviour on Windows.

zerosnacks:
Marking as `not planned`, there is no claim of commands being Windows compatible

PR #1296, opened 2024-09-16
fix: incompatible `-L` flag in `tree lib -L 1` on Windows

zerosnacks:
Closes: https://github.com/foundry-rs/book/issues/1149

The output isn't fully accurate now but good enough to get the point across

### DIFF
--- a/src/output/deps/tree
+++ b/src/output/deps/tree
@@ -1,6 +1,6 @@
 // ANCHOR: all
 // ANCHOR: command
-$ tree lib -L 1
+$ tree lib
 // ANCHOR_END: command
 // ANCHOR: output
 lib


onbjerg (approved):

DaniPopes:
the output directory is autogenerated as you can see from the github diff... @zerosnacks

zerosnacks:
Triggered a re-run of the output directory to revert this change here: https://github.com/foundry-rs/book/pull/1302

I don't see a good way to resolve https://github.com/foundry-rs/book/issues/1149 considering the `tree` command doesn't have compatible flags on the different platforms and we don't make a claim of the commands being runnable on Windows

Going to mark the original ticket as `not planned`

DaniPopes:
Yeah that's fine

PR #1325, opened 2024-10-04
Document `vm.mockCalls`

Tudmotu:
Documentation for `vm.mockCalls`, see https://github.com/foundry-rs/foundry/pull/9024.

### DIFF
--- a/src/cheatcodes/README.md
+++ b/src/cheatcodes/README.md
@@ -307,6 +307,13 @@ interface CheatCodes {
     // function will be mocked.
     function mockCall(address, bytes calldata, bytes calldata) external;
 
+    // Mocks multiple call to an address, returning specified data for each call.
+    //
+    // Calldata can either be strict or a partial match, e.g. if you only
+    // pass a Solidity selector to the expected calldata, then the entire Solidity
+    // function will be mocked.
+    function mockCalls(address, bytes calldata, bytes[] calldata) external;
+
     // Reverts a call to an address, returning the specified error
     //
     // Calldata can either be strict or a partial match, e.g. if you only
--- a/src/cheatcodes/README.md
+++ b/src/cheatcodes/README.md
@@ -307,12 +307,19 @@ interface CheatCodes {
     // function will be mocked.
     function mockCall(address, bytes calldata, bytes calldata) external;
 
+    /// Mocks a call to an address with a specific `msg.value`, returning specified data.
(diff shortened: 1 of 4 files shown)


src/cheatcodes/README.md
```diff
@@ -307,6 +307,13 @@ interface CheatCodes {
     // function will be mocked.
     function mockCall(address, bytes calldata, bytes calldata) external;
 
+    // Mocks multiple call to an address, returning specified data for each call.
+    //
+    // Calldata can either be strict or a partial match, e.g. if you only
+    // pass a Solidity selector to the expected calldata, then the entire Solidity
+    // function will be mocked.
+    function mockCalls(address, bytes calldata, bytes[] calldata) external;
```
Evalir, line 315: the version with `uint256 msgValue` should also be added, no?

src/cheatcodes/mock-calls.md
```diff
@@ -0,0 +1,52 @@
...
+See [`mockCall`](./mock-call.md) for more information on mocking calls and
+matching precedence.
+
+> ℹ️ **Note**
+>
+> Any invocation of the mocked call beyond the number of elements in `retdata`
+> will receive the last `retdata` element in response. `clearMockedCalls` can be
+> called to clear the mock
+
+### Examples
```
Evalir, line 33: We should also have an example that uses the version with `uint256 msgValue`

src/cheatcodes/README.md
Tudmotu: I was not sure, since the `mockCall` with `msgValue` is not included. By I can include it if you think it should be documented. In that case I should probably add it for `mockCall` as well?

src/cheatcodes/mock-calls.md
Tudmotu: I will add and update you

src/cheatcodes/README.md
Tudmotu: Added both `mockCalls` and `mockCalls` with msg.value, lmk if is ok.

src/cheatcodes/mock-calls.md
Tudmotu: Added an example with `msg.value`

Evalir (approved):

Evalir:
deferring to @mattsse / @DaniPopes for merge.

mattsse (approved):

PR #1331, opened 2024-10-11
Fix: Remove Counter.s.sol to resolve forge build error

TomasCImach:
This PR updates the setup instructions to remove the `Counter.s.sol` script file along with other Counter-related files. This change resolves a `forge build` error caused by the presence of the unused Counter script.

Changes made:
- Updated the shell command to remove `script/Counter.s.sol`
- This ensures a clean project setup for the NFT tutorial


DaniPopes (approved):
Thanks!

PR #1334, opened 2024-10-15
doc: add info for vyper lib configuration

Philogy:


src/config/vyper.md
```diff
@@ -12,6 +12,34 @@ Otherwise, you can set the path to `vyper` in your `foundry.toml` by adding the
...
+
+If you want an import like the following to work in your Vyper contract:
+
+```vyper
+from snekmate.utils import eip712_domain_separator
+```
+
+You can install Vyper the desired library via `forge install` e.g. `forge install pcaversaccio/snekmate`.
+
+You then need to adjust your `foundry.toml` as follows (replacing "snemkate" with the name of your
```
Evalir, line 25:
```suggestion
You then need to adjust your `foundry.toml` as follows (replacing "snekmate" with the name of your
```

src/config/vyper.md
```diff
@@ -12,6 +12,34 @@ Otherwise, you can set the path to `vyper` in your `foundry.toml` by adding the
...
+
+```toml
+skip = ["**/lib/snekmate/**"]
+libs = ["lib", "lib/snekmate/src"]
+```
+
+#### Vyper libraries via `pip`
+
+Alternatively if you want to install the package via `pip` into your system's python configuration
+or a virtual environment you can point foundry to it by modifying your `foundry.toml` as follows:
```
Evalir, line 36: Q: Have you tried [`uv`](https://github.com/astral-sh/uv)? Does it work? If so, I think we should also include that. It's much better / faster than `pip`.

Philogy: Yes this works with `uv`, thought it was kind of implicit since generally environments set up with `uv` behave pretty much like normal python.

Philogy: Noted explicitly 👌

Evalir (approved):
awesome, thanks!

onbjerg (approved):

PR #1337, opened 2024-10-22
Update default `out` folder

guidanoli:
This should be checked/generated automatically.


onbjerg (approved):

DaniPopes:
No this is not right

https://github.com/foundry-rs/foundry/blob/25598999a2b33ac6ccfa35c347f3c98aba8e0061/crates/config/src/lib.rs#L2155

PR #1403, opened 2025-01-28
Update expect revert sigs, add warning re expect revert depth and next call

### LINKED ISSUE #1374: vm.expectRevert() does not necessarily bind to the primary function
As an example the following lines:
```
vm.expectRevert();
stable.donate(sUSD.balanceOf(users[1]));
```
the ```stable.donate()``` function is indeed expected to revert, but since ```sUSD.balanceOf(users[1])``` does succeed the ```expectRevert()``` will make the test fail, with ```[FAIL: next call did not revert as expected]```.

An easy workaround is though to simply rewrite the order:

```
uint256 susdToDonate = sUSD.balanceOf(users[1]);
vm.expectRevert();
stable.donate(susdToDonate);
```

I'm unsure if this is expected behavior, but I find it quite unintuitive.

grandizzy:
Closes #1374

### DIFF
--- a/src/cheatcodes/expect-revert.md
+++ b/src/cheatcodes/expect-revert.md
@@ -11,29 +11,79 @@ function expectRevert(bytes4 message) external;
 ```
 
 ```solidity
+function expectRevert(bytes4 message, address reverter) external;
+```
+
+```solidity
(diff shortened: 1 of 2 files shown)


zerosnacks (approved):
:tada:

Issue #1440, opened 2025-02-07
Document `cast wallet sign-auth` and limitations around nonce when signing and sending


zerosnacks:
Document cast wallet sign-auth and clarify use of cast send directly

PR #1441, opened 2025-02-07
chore: document `cast send --auth` and `cast wallet sign-auth` (EIP-7702)

zerosnacks:
Closes: https://github.com/foundry-rs/book/issues/1440

Looks like the references have gotten a bit incomplete, will look into how we can more easily automate this.

Added the note on nonce management when using `sign-auth` when intending to send with the same signer


mattsse (approved):
lgtm, pending @klkvr

klkvr (approved):
nice

PR #1453, opened 2025-02-17
Additional compilers docs

grandizzy:
- add additional compilers docs
- update v1.0 migration guide to point to missing `optimizer` and additional compilers usage
- update v1.0 migration guide to warn re usage of new console.log with older signatures / need to update forge-std


klkvr (approved):
lgtm!

PR #1456, opened 2025-02-20
chore(`cast`): rename `4byte-decode` -> `4byte-calldata`, old remains as hidden alias

zerosnacks:
Related: https://github.com/foundry-rs/foundry/pull/9927


grandizzy (approved):
lgtm

aspiers:
I'm on the latest stable version of `cast`, but `4byte-calldata` isn't yet a valid command:

```
$ cast --version
cast Version: 1.0.0-stable
Commit SHA: e144b82070619b6e10485c38734b4d4d45aebe04
Build Timestamp: 2025-02-13T20:03:31.026474817Z (1739477011)
Build Profile: maxperf
$ cast 4byte-calldata   
error: unrecognized subcommand '4byte-calldata'

  tip: some similar subcommands exist: '4byte-decode', '4byte-event', 'ta', 'ca', '4b', '4e', 'ba', 'decode-calldata', 'calldata', '4be', 'call', '4byte', 'pretty-calldata'

Usage: cast [OPTIONS] <COMMAND>

For more information, try '--help'.
```

I'm guessing that this is because the stable release is lagging behind both the latest code and documentation in GitHub.  But this is quite confusing for normal users such as myself who don't live on the bleeding edge and would naturally expect the latest stable release to match the official docs at https://book.getfoundry.sh/reference/cli/cast

Is there a way to solve this, e.g. by having separate builds published on separate URLs for the stable version of the docs and the latest version in git?

zerosnacks:
> Is there a way to solve this, e.g. by having separate builds published on separate URLs for the stable version of the docs and the latest version in git?

Noted - that is a fair point! It is something we have considered but hasn't been picked up yet due to lack of bandwidth. We are still experimenting with the speed at which stable versions are released, hoping that shortening the cycle will make this less of an issue.

aspiers:
Thanks a lot for the very helpful and quick response!

PR #1469, opened 2025-03-09
fix: mention `always_use_create_2_factory` in guide

### LINKED ISSUE #1462: Mention `always_use_create_2_factory` option in guides/deterministic-deployments-using-create2
In this section (and in the whole guide), there is no mention of the `always_use_create_2_factory` option.

https://github.com/foundry-rs/book/blob/90aba43048764c45d5b81aafdfaf2d368c84e2ec/src/guides/deterministic-deployments-using-create2.md?plain=1#L23-L33

My understanding is that, by default, create2 deployments will not work in tests and scripts that don't broadcast (https://github.com/foundry-rs/foundry/issues/6402). I think it would be practical to at least mention it in the guide.

For context, I spent a few hours thinking I was doing something wrong when seeing that my contract was not deployed at the right address in my tests. Including `always_use_create_2_factory = true` in `foundry.toml` fixed it.

I can propose a pull request if this issue is accepted.

Pybast:
fix #1462

### DIFF
--- a/src/guides/deterministic-deployments-using-create2.md
+++ b/src/guides/deterministic-deployments-using-create2.md
@@ -32,9 +32,15 @@ bytecode_hash = "none"
 cbor_metadata = false
 ```
 
+### Solc version
+
 It is required to pin your `solc` (Solidity) version. It is generally recommended to use a recent version or, if preferred, the latest version.
 
+### EVM version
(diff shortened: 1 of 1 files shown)


mattsse (approved):
ty!

PR #1477, opened 2025-03-17
fix(`debugger`): incorrect docs after removal of `forge debug`

zerosnacks:
Related: https://github.com/foundry-rs/foundry/issues/10079

This doc was previously written for the standalone `debug` command, now that it is rolled into `forge test` and `forge script` the default of `<PATH>` is not very helpful in the context of function debugging. Update the doc to point the user to use `--match-test` to select a single function instead.

### DIFF
--- a/src/forge/debugger.md
+++ b/src/forge/debugger.md
@@ -2,23 +2,25 @@
 
 Forge ships with an interactive debugger.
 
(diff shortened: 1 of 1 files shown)


grandizzy (approved):
lgtm!

PR #1480, opened 2025-03-18
Add --no-match-coverage flag to coverage command docs

ScreamingHawk:
Added flag as per this change: https://github.com/foundry-rs/foundry/pull/8321


zerosnacks:
Thanks!

PR #1481, opened 2025-03-19
 Fix incorrect return variable in getBlockNumber documentation

PixelPil0t1:
Fixed return variable name:

Old: returns (uint256 timestamp);
New: returns (uint256 blockNumber);
Reason: Function returns block.number, not a timestamp.

### DIFF
--- a/src/cheatcodes/get-block-number.md
+++ b/src/cheatcodes/get-block-number.md
@@ -3,7 +3,7 @@
 ### Signature
 
 ```solidity
-function getBlockNumber() external view returns (uint256 timestamp);
+function getBlockNumber() external view returns (uint256 blockNumber);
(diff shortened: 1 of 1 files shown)


src/cheatcodes/get-block-number.md
```diff
@@ -3,7 +3,7 @@
 ### Signature
 
 ```solidity
-function getBlockNumber() external view returns (uint256 timestamp);
+function getBlockNumber() external view returns (uint256 blockNumber);
```
zerosnacks, line 6:
```suggestion
function getBlockNumber() external view returns (uint256 height);
```

zerosnacks (approved):
Thanks!

Issue #1483, opened 2025-03-19
Update cast ABI Commands part (outdated and missing decode-XXX commands)


Hugoo:
Following this PR: https://github.com/foundry-rs/foundry/pull/9413

which renamed decoding functions to: `decode-XXX`

The CLI reference is ok:

https://book.getfoundry.sh/reference/cli/cast/decode-abi

<img width="260" alt="Image" src="https://github.com/user-attachments/assets/b677d83c-bf2d-4d4a-a079-e7087bf53e49" />

But the "cast ABI Commands" part is outdated and missing some functions (`cast decode-error`, etc.)

https://book.getfoundry.sh/reference/cast/cast-abi-decode

<img width="241" alt="Image" src="https://github.com/user-attachments/assets/22cec2d1-d757-43ce-865a-19772c21b45a" />

Maybe creating a "Decoding" section could also make sense.

PR #1484, opened 2025-03-20
Update abi/calldata-decode to decode-abi/calldata

Hugoo:
Related to: https://github.com/foundry-rs/book/issues/1483

### Note

- I usually like to set a redirection between the old page and the new page to preserve SEO -> lmk if you want me to add it
- There are still missing commands, see the related issue. I can handle it lmk - I didn't want to put too much effort before having your feedback. However, the PR can be merged as it is.


mattsse (approved):
great, ty.

while these aliases still work, we should be consistent in the docs

PR #1487, opened 2025-03-26
It's `fs_permissions` not, `fs_permission`

DanielVF:
Be me. Spend 30 minutes trying to figure out why I couldn't get permissions to write a csv file, when I had copied the config name out of the docs. It's `fs_permissions` not, `fs_permission`.

Fix the docs, for those who follow after.


zerosnacks (approved):
Thanks!

PR #1501, opened 2025-04-22
debugger - added cast run info and example

radeksvarz:
Debugger can be run on already executed transactions. Adding missing info.


zerosnacks:
Thanks!

zerosnacks (approved):

PR #1530, opened 2025-05-30
fix(`landing`): increase foundry banner size

yash-atreya:
Redo top section of landing page to highlight mascot.


yash-atreya:
Full width might be a bit over the top. 

Users would have to scroll to click get-started / copy the installation instructions.

Preview URL: https://1aaee359.foundry-book.pages.dev/

Screenshot 2025-05-30 at 7 07 00 PM

zerosnacks:
Per https://github.com/foundry-rs/book/pull/1537

I think it would be nice if we can feature a link to the announcements on the homepage as we had

PR #1566, opened 2025-06-11
function name signatures in expect-revert.mdx are inaccurate

L0GYKAL:
Remove unexisting "expectRevert" signatures
The deleted signatures just don't exist so they shouldn't be in the documentation

### DIFF
--- a/vocs/docs/pages/reference/cheatcodes/expect-revert.mdx
+++ b/vocs/docs/pages/reference/cheatcodes/expect-revert.mdx
@@ -7,19 +7,55 @@ function expectRevert() external;
 ```
 
 ```solidity
-function expectRevert(bytes4 message) external;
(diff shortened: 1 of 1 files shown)


L0GYKAL:
ok I'll do the naming update but if they still exist I don't understand why it wouldn't compile on my end =(
I tried them all

zerosnacks:
> ok I'll do the naming update but if they still exist I don't understand why it wouldn't compile on my end =( I tried them all

what version of `forge-std` are you using? you may need to update it using `forge update`

L0GYKAL:
Right, it was 1.0, after forge update it's now 1.9.7
Thanks!

zerosnacks (approved):

zerosnacks:
Thanks! Glad it also got resolved on your end

PR #1590, opened 2025-07-03
Add Chainstack faucet

akegaviar:


zerosnacks (approved):
Thanks, I've reorded the faucets based on the amount of ETH they require to hold on mainnet, from none for POW faucet to 0.08 ETH on Chainstack

PR #1593, opened 2025-07-07
docs(forge): EIP-7907, document Osaka evm-version and ignore-initcode-size in build

jgresham:


mattsse (approved):
lgtm

pending @zerosnacks

zerosnacks (approved):
Makes sense, the additional included commits due to a force push that happened on `master` should not create issues.

PR #1601, opened 2025-07-19
Update deploying.md

Rahul-cyber26:
#1600


zerosnacks:
Users need to get their own API keys per Etherscans terms of service

PR #1635, opened 2025-08-15
docs(anvil): add troubleshooting note for chained txs when automine i…

anil2ec4:
### Summary
Adds a troubleshooting note explaining why a second tx can be rejected with
“Insufficient funds for gas * price + value” when `automine` is disabled, and
lists two workarounds. Docs-only change; no code.

### Context
Follow-up to foundry-rs/foundry#11264 — moved here per maintainer guidance.

### DIFF
--- a/vocs/docs/pages/anvil/reference/anvil.mdx
+++ b/vocs/docs/pages/anvil/reference/anvil.mdx
@@ -361,3 +361,15 @@ EVM options:
           
           [aliases: --tracing]
 ```
+Troubleshooting: Chained transactions with automine disabled
+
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
this is useful, can we extend this with this new cli arg here, which would bypass this rule

https://github.com/foundry-rs/foundry/pull/11242/files#diff-821677e726ecf3de9d59a8e9cc1aef2c64650c7b04452be6860ed123ad29e17aR596-R598

anil2ec4:
Great point — added an “Advanced” workaround for `--disable-pool-balance-checks` with a short note and example.

Thanks!

mattsse (approved):

PR #1642, opened 2025-08-25
chore: add note on disabling linting on build 

zerosnacks:
cc @0xrusowsky

As asked here: https://x.com/siobh_eth/status/1959988063003689201


grandizzy (approved):
👍

PR #1652, opened 2025-09-09
Improve JSON path documentation for writeJson and writeToml cheatcodes

### LINKED ISSUE #1638: Document JSON paths more clearly
The docs for JSON paths for some cheat codes like `vm.writeToml` and `vm.writeJson` are not very clear. We should include some more information about how the JSON path is expected to be formed. For example, using `"key.foo"` will fail saying the path doesn't exist, because it silently gets changed to `"$key.foo"` - what the user should really do here is either `".key.foo"` or `"$.key.foo"`.

The JSON paths are from https://docs.rs/jsonpath_lib/latest/jsonpath_lib/

Signor1:
Fixes #1638

Clarifies JSON path syntax in the documentation for `vm.writeJson` and `vm.writeToml` cheatcodes.

**Changes:**
- Explains that paths like `"key.foo"` get silently converted to `"$key.foo"` and fail
- Documents that paths should start with `.` or `$` (e.g., `".key.foo"` or `"$.key.foo"`)
- Adds reference to the underlying jsonpath_lib library
- Includes examples of valid path formats

This addresses user confusion about why certain JSON paths don't work as expected.

### DIFF
--- a/vocs/docs/pages/reference/cheatcodes/write-json.mdx
+++ b/vocs/docs/pages/reference/cheatcodes/write-json.mdx
@@ -50,9 +50,22 @@ Let's consider the following JSON object:
 }
 ```
 
-The root object is always assumed, so we can refer to one of its children by starting the path with a dot (`.`). For instance, `.boolean`, `.number`, and `.obj1`.
-We can go as deep as we want: `.obj1.aNumber`, or `.obj1.obj2.aNumber`.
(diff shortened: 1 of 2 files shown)


zerosnacks:
Thanks! Changes make sense, pending review of @onbjerg

Signor1:
@all-contributors please add @Signor1 for Maintenance

PR #1694, opened 2025-11-04
Update tree command to show two directory levels

jormarma:
After running the updated command, you can now see the `solady` library under the `lib` folder. Before this change, it wasn’t showing up.


zerosnacks:
Ref: makes sense https://github.com/foundry-rs/book/blob/e9b74f99ebf9e810fc18bf87aee0b0b85b013c06/vocs/docs/pages/guides/project-setup/dependencies.md?plain=1#L21-L23

zerosnacks:
Thanks!

PR #1698, opened 2025-11-10
docs(anvil): add supported Beacon Node API endpoints

mablr:
List of supported Beacon Node API endpoints, with their description and link to specification.


mattsse (approved):
nice

PR #1755, opened 2026-02-03
feat: add terminal directive for help

tmm:
Adds [`:::terminal`](https://vocs-next.vercel.app/guide/syntax-highlighting#terminal-directive) to autogenerated help files


onbjerg (changes requested):
can you regen these as well

PR #1785, opened 2026-04-14
docs(scripting): clarify the usage of --rpc-url vs that of --chain

### LINKED ISSUE #12907: bug: documentation --chain <chain-name> fails during script usage
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

forge Version: 1.5.0-stable Commit SHA: 1c57854462289b2e71ee7654cd6666217ed86ffd Build Timestamp: 2025-11-26T09:16:58.269730000Z (1764148618) Build Profile: maxper

### What version of Foundryup are you on?

foundryup: 1.4.0

### What command(s) is the bug in?

forge script

### Operating System

macOS (Apple Silicon)

### Describe the bug

In the documentation here: https://getfoundry.sh/guides/scripting-with-solidity#deploying-to-a-testnet it says uses "--chain <chain-name" based on foundry.toml config but it actually requires using --rpc-url

Unclear if this is a bug in the CLI or if the docs should be updated. It would be nice if both worked.

snehendu098:
Closes: [foundry-rs/foundry#12907](https://github.com/foundry-rs/foundry/issues/12907)

Users reported that the scripting guide implied `--chain <chain-name>` could select a network for `forge script`, but in practice `--rpc-url` is required. `--chain` only sets `block.chainid` in the EVM env — it does not pick an RPC endpoint.

The old misleading example was removed in the Feb reorg, but the current `forge/scripting.mdx` never explicitly showed the `[rpc_endpoints]` alias pattern, so the same confusion can recur.

## Changes

I have added a small subsection to `src/pages/forge/scripting.mdx` which shows the usage of `[rpc_endpoints]` in `foundry.toml` and also shows `forge script ..... --rpc-url sepolia` using the alias

I have also added a note clarifying that `--chain` only selects a `block.chainid` and doesn't select an RPC

## Verification
I had initialized a new project using `forge init` which gives a basic counter app and made a change in the `foundry.toml` file

```toml
[rpc_endpoints]
sepolia = "https://ethereum-sepolia-rpc.publicnode.com"
```

- `forge script script/Counter.s.sol --rpc-url sepolia` → resolves alias, simulates against Sepolia (chain id `11155111`), writes to `broadcast/Counter.s.sol/11155111/dry-run/`.
- `forge script script/Counter.s.sol --chain sepolia --broadcast` → runs locally against an empty EVM, no RPC resolution, prints *"If you wish to simulate on-chain transactions pass a RPC URL."*

Confirms `ScriptArgs` impl is correct; fix is docs-only, matching @mablr's earlier comment on the issue.


mattsse (approved):
Docs-only clarification looks good after dropping the unrelated generated artifacts.

PR #1826, opened 2026-05-09
docs: document --no-lint for forge build

stevencartavia:
Document the new `forge build --no-lint` flag added in https://github.com/foundry-rs/foundry/pull/14676, alongside the existing `lint_on_build = false` config option.


mattsse (approved):
Docs-only update for build lint flags looks good.

PR #1848, opened 2026-05-15
feat(lint): add `assert-state-change`

mablr:
cf. https://github.com/foundry-rs/foundry/pull/14784


mattsse (approved):
Docs look good from this pass.

PR #1849, opened 2026-05-15
feat(lint): add `unused-return`

mablr:
cf. https://github.com/foundry-rs/foundry/pull/14785


mattsse (approved):
Docs look good from this pass.

PR #1868, opened 2026-05-23
docs: add unprotected initializer lint page

figtracer:
## Summary
- add the book page for the new `unprotected-initializer` lint
- link it from the linting index and high-severity sidebar

Pairs with foundry-rs/foundry#14886.

## Tests
- `bun run build`

### DIFF
--- a/src/pages/forge/linting/unprotected-initializer.mdx
+++ b/src/pages/forge/linting/unprotected-initializer.mdx
@@ -16,7 +16,7 @@ Warns on public or external functions that:
 
 1. are marked with an `initializer` or `reinitializer` modifier;
 2. write state directly or through an internal helper;
-3. belong to an implementation whose constructor does not call `_disableInitializers()`; and
+3. belong to an implementation whose constructor does not call an inherited `_disableInitializers()`; and
 4. are in a contract with a public or external `delegatecall`, `callcode`, or `selfdestruct` path
    that is not restricted to proxy calls with `onlyProxy`.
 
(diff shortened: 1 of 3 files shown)


src/pages/forge/linting/unprotected-initializer.mdx
```diff
@@ -0,0 +1,74 @@
...
+an implementation contract that exposes a destructive entry point.
+
+### What it does
+
+Warns on public or external functions that:
+
+1. are marked with an `initializer` or `reinitializer` modifier;
+2. write state directly or through an internal helper;
+3. belong to an implementation whose constructor does not call `_disableInitializers()` and is not
+   itself marked `initializer`; and
```
mattsse: This should not document `constructor() initializer` as equivalent to `_disableInitializers()`. The constructor modifier locks the initial version, but does not necessarily prevent later `reinitializer(n)` entrypoints from being called directly on the implementation.

```suggestion
3. belong to an implementation whose constructor does not call `_disableInitializers()`; and
```

mattsse (changes requested):
This needs to stay in sync with foundry-rs/foundry#14886. The current page documents `constructor() initializer` as a full implementation lock, but that is the unsafe exemption we should remove: it only initializes/locks version 1, while later `reinitializer(n)` functions can still be callable on the implementation.

mablr (commented):
This should be included https://github.com/foundry-rs/foundry/pull/14886/commits/9cc4336f7ad177d8e579f0876b19a95d383f2122

mattsse (approved):
LGTM. The book page now matches the Foundry lint docs wording and the previous initializer-lock concern is resolved.

PR #1869, opened 2026-05-23
chore(lint): resync `event-fields`, `uninitialized-state`, & `weak-prng`

mablr:


mattsse (approved):
The lint limitation updates read consistently with the current behavior and improve the docs. No content blockers found; the Vercel failure appears to be the preview integration/auth check rather than a docs build issue.

PR #1875, opened 2026-05-26
docs(lint): add `write-after-write`

mablr:
cf. https://github.com/foundry-rs/foundry/pull/14915


mattsse (approved):
lgtm

mattsse (commented):
Docs look consistent with the intended conservative scope, especially the “flat sequence” wording here: [`write-after-write.mdx#L16-L20`](https://github.com/foundry-rs/book/blob/56186faf690cd0f2d9ad8c0954ebe2a5e8de443e/src/pages/forge/linting/write-after-write.mdx#L16-L20). I would keep this as-is if the foundry implementation is fixed to respect expression-level branching/short-circuiting. If the implementation intentionally keeps scanning ternary arms and short-circuit operands linearly, this line should be updated before merge.

PR #1876, opened 2026-05-26
docs(lint): add `encode-packed-collision`

mablr:
cf. https://github.com/foundry-rs/foundry/pull/14914

### DIFF
--- a/sidebar/sidebar.ts
+++ b/sidebar/sidebar.ts
@@ -45,6 +45,7 @@ const docs = [
             collapsed: true,
             items: [
               { text: "arbitrary-send-erc20", link: "/forge/linting/arbitrary-send-erc20" },
+              { text: "encode-packed-collision", link: "/forge/linting/encode-packed-collision" },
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

mattsse (commented):
Non-blocking if foundry-rs/foundry#14914 fixes this before release, but the docs currently make the broad claim that the lint flags `abi.encodePacked()` with two or more dynamic args: [`encode-packed-collision.mdx#L17-L22`](https://github.com/foundry-rs/book/blob/d2388e159b88e9202ad2fc5dc51da95231443884/src/pages/forge/linting/encode-packed-collision.mdx#L17-L22).

That claim conflicts with the current foundry implementation’s false negative for overridden inherited member calls: [`encode_packed_collision.rs#L110-L126`](https://github.com/foundry-rs/foundry/blob/19d7dbd3b4d164bbd2309b21618a1bdeb889414d/crates/lint/src/sol/high/encode_packed_collision.rs#L110-L126).

Suggested fix: keep the docs as-is if the linter is fixed, or add a temporary limitation note for member calls on concrete contracts with overridden inherited methods.

PR #1918, opened 2026-06-29
docs: expand Tempo guide

mattsse:
This rewrites the Tempo guide as a Foundry-native hub instead of a short setup note. It covers current setup and hardfork guidance, Tempo transaction capabilities, TIP-20 and virtual-address workflows, TIP-403 and receive-policy tooling, access keys and keychain workflows, and local testing/debugging paths. The guides index card now reflects that broader scope.

The guide keeps generated command references as the source of truth for flag-level syntax and links protocol semantics out to Tempo docs or TIP pages. It also removes stale historical-hardfork-first examples so normal setup points at `network = "tempo"`, live RPC inference, and current T6 behavior.

Resolves Linear OSS-441, OSS-442, OSS-443, OSS-444, OSS-445, OSS-446, and OSS-447.

AI assistance was used to draft this documentation update.

PR #1919, opened 2026-06-29
docs: update Tempo setup guidance

mablr:
This updates the Tempo setup guidance to make upstream Foundry the normal supported path, clarify `forge init --network tempo`, and explain when `network = "tempo"` is needed versus when live Tempo RPC forks can infer network semantics.

Hardfork pinning is now presented as explicit opt-in behavior for deterministic activation-point testing, using current T6 behavior as the example.

Closes OSS-441.


mattsse (approved):

mablr (commented):
Rebased on #1918. Merging the missing points

PR #1924, opened 2026-07-02
docs(lint): add unused-error page

0xMars42:
Adds the Foundry Book page for the `unused-error` lint and wires it into the linting index and sidebar under Informational.

This gives `https://getfoundry.sh/forge/linting/unused-error` a published page for the lint added in foundry-rs/foundry#15539 (from the parity checklist foundry-rs/foundry#14381), so the Foundry `ensure_lint_rule_docs` check has matching book coverage.


mattsse (approved):
Looks good. The new lint page matches the implementation docs and the local Vocs build succeeds.

PR #1969, opened 2026-07-15
docs(anvil): add mining and pool guide

mattsse:
Adds a conceptual Anvil guide for instant, interval, mixed, and manual mining, including runtime mode changes, timestamp control, transaction ordering, pool inspection, and pending-transaction removal. It also documents a deterministic automation workflow based on RPC readiness, asynchronous submission, explicit block production, and structured node and pool responses so agents do not need to scrape terminal logs.

### DIFF
--- a/src/pages/anvil/mining.mdx
+++ b/src/pages/anvil/mining.mdx
@@ -1,0 +1,162 @@
+---
+description: Control Anvil mining, inspect pending transactions, and build deterministic local workflows.
+---
+
+## Mining and Transaction Pool
+
+Anvil can mine immediately, on a fixed interval, both ways, or only when you request a block. Choose the mode that matches the behavior you need to test instead of relying on the default instant-mining behavior.
+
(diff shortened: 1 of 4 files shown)


src/pages/anvil/mining.mdx
```diff
@@ -0,0 +1,162 @@
...
+```
+
+Inspect the current mode before changing it:
+
+```bash
+$ cast rpc anvil_getAutomine
+$ cast rpc anvil_getIntervalMining
+```
+
+These RPC methods replace the current mining mode; they do not preserve mixed mode. Prefer startup flags when a long-running node must stay in mixed mode.
```
mablr: evm_setAutomine false does not always replace the current mode: it is a no-op in interval or mixed mode. The two getters also cannot identify mixed mode because they report false and no interval. Could we document these limitations instead of saying they inspect and replace the current mode?

mablr (approved):

PR #1975, opened 2026-07-15
docs(reference): add generation provenance

mattsse:
Makes generated CLI reference pages attributable and searchable. The help generator now emits frontmatter descriptions for every command, a concise source marker on every page, and a visible provenance note on each root tool reference. It also generates one versions page containing the exact Forge, Cast, Anvil, and Chisel binaries, avoiding version churn across hundreds of files while giving people and agents a canonical place to resolve nightly drift. The reference files in this pull request were regenerated without substantive command-help changes.


mablr (approved):
mechanical, lgtm

PR #1976, opened 2026-07-15
ci: check cheatcode reference coverage

mattsse:
Adds a deterministic cheatcode coverage manifest and a Bun check that compares the canonical Foundry function surface with authored pages and sidebar routes. The reviewed baseline records 313 non-internal function names, of which 191 are currently mentioned and 122 remain known gaps; CI now fails on new undocumented names, stale gap entries, broken sidebar routes, or orphan pages. The weekly update job refreshes the manifest from the exact commit reported by the installed nightly binary, so upstream additions become actionable failures instead of silently missing from the book.


scripts/check-cheatcode-coverage.ts
```diff
@@ -0,0 +1,137 @@
...
+  };
+}
+
+const docFiles = readdirSync(docsDir)
+  .filter((file) => file.endsWith(".mdx"))
+  .sort();
+const docs = docFiles
+  .map((file) => readFileSync(join(docsDir, file), "utf8"))
+  .join("\n");
+
```
mablr, line 77: Could we require an actual cheatcode reference, such as `vm.<name>` or a declared signature, instead of any bare occurrence? This currently counts `contains` from ordinary prose in `ffi.mdx` and `replace` from unrelated JSONPath prose, so undocumented additions with ordinary-word names can pass the coverage check.

mablr (approved):

PR #1978, opened 2026-07-15
docs: scope lint rule navigation

mattsse:
The shared conceptual sidebar currently nests all 86 lint-rule pages under Forge, so the full catalog appears while browsing unrelated tools and ahead of practical guides in the generated agent index. This keeps Linting as a normal Forge destination and gives `/forge/linting` a focused sidebar containing the same severity groups and every existing rule.

The Forge entries are shared between both views to prevent navigation drift. Separating the catalog also places task-oriented documentation and guides before individual lint rules in `llms.txt`.


sidebar/forge.ts
```diff
@@ -0,0 +1,121 @@
+import type { SidebarItem } from "./types";
+
+export const forgeDocs: SidebarItem[] = [
+  { text: "Overview", link: "/forge" },
+  { text: "Building", link: "/forge/build" },
+  { text: "Testing", link: "/forge/testing" },
```
mablr, line 8: This extracted list is already stale: current `master` includes `/forge/contract-bindings`, but `forgeDocs` omits it, so resolving the conflict by keeping this array would remove that page from Forge navigation. Could we rebase and preserve the current Forge entries?

mablr (changes requested):

mablr (approved):

PR #1983, opened 2026-07-15
docs(lint): add reentrancy unlimited gas

mattsse:
This documents the `reentrancy-unlimited-gas` lint added in foundry-rs/foundry#15768, explains its path-sensitive built-in `transfer` and `send` analysis and boundaries, removes the obsolete redirect to `reentrancy-eth`, and adds the rule to the info-severity reference index and sidebar.

PR #1985, opened 2026-07-15
docs(lint): add reentrancy unlimited gas

mattsse:
This documents the `reentrancy-unlimited-gas` lint added in foundry-rs/foundry#15768, explains its path-sensitive built-in `transfer` and `send` analysis and boundaries, removes the obsolete redirect to `reentrancy-eth`, and adds the rule to the info-severity reference index and sidebar.


mattsse:
Closing because the original companion PR #1983 was merged while this replacement was being created.

PR #1986, opened 2026-07-15
docs(lint): add reentrancy balance

mattsse:
This documents the `reentrancy-balance` lint added in foundry-rs/foundry#15769, explains its stale native-balance analysis and exclusions, and adds the rule to the high-severity reference index and sidebar.


mattsse:
Closing because the original companion PR #1984 was merged while this replacement was being created.

PR #2012, opened 2026-07-26
docs(forge): document debugger features and networks config

mattsse:
The interactive debugger gained a storage explorer, transient storage inspection, interactive test selection, jump-to-source, pane toggling, and constructor argument display over the last months with no documentation. This rewrites the debugger section of the debugging page around the current TUI: the five panes, the real key bindings (the previous table listed keys like n/s/o/c that do not exist), the command prompt including :storage/:transient/:line/:pc and pane toggles, layout control via l and --debug-layout, and the additional entry points cast run --debug and cast call --trace --debug. The stale key table in the debugging-transactions guide is replaced with a link to the reference so the two pages cannot drift apart again.

The scripting page now covers the --sender-nonce override, and the networks section of the config tooling reference was corrected and expanded: network selects the ethereum/optimism/tempo family, celo is a real (not legacy) toggle for the Celo transfer precompile, and bypass_prevrandao substitutes a random prevrandao when a forked block lacks one rather than being Celo-specific.

All key bindings, commands, and behavior claims were verified against the debugger source rather than the shipped PR descriptions.

### DIFF
--- a/src/pages/config/reference/tooling.mdx
+++ b/src/pages/config/reference/tooling.mdx
@@ -44,9 +44,29 @@ the [Soldeer guide](/projects/soldeer) for its fields and workflows.
 
 ### Networks
 
-The `network` field selects a network family (`ethereum`, `optimism`, or
-`tempo`). Foundry also accepts the legacy `celo` and `tempo` boolean fields.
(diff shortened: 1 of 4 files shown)


mablr (approved):
👌

PR #2021, opened 2026-07-27
ci: refresh coverage baselines after docs merges

mattsse:
The coverage checks from #2014 merged before the docs PRs that fix their baselined violations (#2010, #2011, #2013, #2016), so master currently fails both checks on stale entries — the fail-on-stale semantics doing their job, just in the wrong merge order.

This refreshes both reviewed lists and goes one step further: the three violations that survived #2016 are fixed instead of re-baselined (cast 4byte-decode was renamed upstream to 4byte-calldata in the cast index and FAQ, and the abi-encoding page decoded event data with a nonexistent abi-decode --event; the working command is cast decode-event --sig). Result: the anvil knownMissing list and the narrative-command baseline are both empty, so any future violation of either check is a hard CI failure with no grandfathered entries.


mablr (approved):

src/pages/cast/abi-encoding.mdx
```diff
@@ -71,7 +71,7 @@ $ cast sig-event "Transfer(address indexed,address indexed,uint256)"
 ### Decoding events
 
 ```bash [Decode event data]
-$ cast abi-decode --event "Transfer(address indexed from, address indexed to, uint256 value)" $DATA
+$ cast decode-event --sig "Transfer(address indexed from, address indexed to, uint256 value)" $DATA
```
stevencartavia: `cast decode-event --sig` only supplies topic 0, so this indexed `Transfer` signature fails because the `from` and `to` topics are missing. Could we use a non-indexed event such as `ValueChanged(uint256 value)` instead?

mattsse: Good catch — confirmed empirically: the indexed signature fails with `invalid log topic list length: expected 3 topics, got 1`, since `--sig` only supplies topic0 and the data payload. Switched the example to a non-indexed `ValueChanged(uint256 value)` as suggested (verified it decodes).

stevencartavia (approved):

PR #2025, opened 2026-07-29
docs(config): document nested remappings

stevencartavia:
Documents contextual remapping syntax and explains how Foundry scopes nested dependency refinements while retaining package-root fallbacks and explicit remapping precedence. Companion to foundry-rs/foundry#15929. AI assistance was used to draft and review these changes.


mattsse (approved):
lgtm

PR #2039, opened 2026-08-17
docs: fix broken links in Std Assertions index

sleepycat12341013:
## What happened

The 19 links on the Std Assertions index page end in `.mdx`, and they do not
resolve on the published site.

## Evidence

- https://getfoundry.sh/reference/forge-std/assertEq.mdx → 404
- https://getfoundry.sh/reference/forge-std/assertEq → renders correctly

## What this changes

I dropped the extension so each link points at the pages that actually exist.
All 19 targets are present under `src/pages/reference/forge-std/`, and no other
content is changed.

## About the scope

The same pattern appears elsewhere in the book: 121 links across 38 files carry
the `.mdx` extension, against 920 that omit it.

I kept this PR to a single file so it is easy to check. Happy to follow up with
the rest — as one PR, or split by section, whichever you prefer.


mablr (commented):
@sleepycat12341013 do you mean on this page https://www.getfoundry.sh/reference/forge-std/std-assertions ? For me the links seem to work correctly. When I click on any link of the page I'm already correctly redirected. eg [assertEq](https://getfoundry.sh/reference/forge-std/assertEq) -> https://getfoundry.sh/reference/forge-std/assertEq

sleepycat12341013:
You're right that clicking the links on the rendered page works.
but this issue is with the raw URLs.
Could you try copy and paste input field https://getfoundry.sh/reference/forge-std/assertEq.mdx 
This result is 404.
Happy to close it if you'd rather keep it as-is.

onbjerg:
our static site generator makes these links resolve

PR #2056, opened 2026-09-03
feat: add performance site

DaniPopes:
Adds the Solar performance site at `/perf/solar/` alongside the Foundry docs: per-benchmark history graphs, commit comparisons, and dynamically discovered compiler artifacts with split/unified diffs. `/perf/` temporarily redirects there; a TODO tracks replacing it with a generic dashboard.

The dashboard plots the latest 60 main runs separately for each benchmark, preserving missing measurements as gaps. Expanded history loads only the selected benchmark and metric. Compare inputs accept commits, prefixes, branches, tags and PRs without defaults or suggestions. The table shows Solar Head values and relative Base/compiler costs, with compiler versions and raw values on hover. Compiler columns and artifact lists are data-driven, including real solc and solx results from Solar PR #1412. Independent Left/Right selectors let any available compilers be compared without hardcoding their artifact filenames.

Frontend, read API and authenticated import worker deploy together on Vercel. GitHub App imports publish immutable ClickHouse snapshots only after their content-addressed artifact bodies exist. Comparisons and viewer descriptors each batch both runs into one query; manifests are scoped to the selected benchmark. Revision-pinned links, bounded browser/origin caches, content-hash reuse and parallel loading reduce repeat work. Large artifacts use bounded previews and full downloads. The index returns its 12 displayed rows plus totals and predecessor commits.

The active database now uses only `run_snapshots`, `artifact_blobs` and `ingestion_jobs`. Legacy data was migrated and the old tables removed after verifying a retained backup; fresh imports work through the deployed worker after cleanup. Rolling back to pre-snapshot code requires restoring that backup. Setup, SQL timings, migration and rollback are documented in `perf/README.md`.

Missing runs import on demand with bounded polling. The protected production cron runs every 15 minutes; its first scheduled execution still needs verification after rollout. Dispatch deduplication remains per-instance, not a durable distributed queue, and artifact bodies remain in ClickHouse. `PERF_DEMO_DATA=1` provides an explicit credential-free local/demo mode.

The [live preview](https://foundry-docs-git-fork-danipopes-dani-move-perf-site-tempoxyz.vercel.app/perf/solar/) includes historical runs backfilled from August 9 onward. Older runs may have metrics without artifacts; expired GitHub archives cannot be recovered. Screenshots below use real imported data.

AI assistance: implementation and testing with Codex.

Solar per-benchmark history dashboard

Solar, solc and solx benchmark comparison

Solar versus solx compiler artifact diff


DaniPopes:
pending prod vars

mablr (approved):
🚀

figtracer:
when checking different metrics e.g.: "compile time" vs "runtime gas", it should save the one you're analyzing
on reload 

https://foundry-docs-git-fork-danipopes-dani-move-perf-site-tempoxyz.vercel.app/perf/solar/?base=e4dd44725141090b0cd3a7b6a5f286babb111189&head=36ca54c2220901408b18123746d31dfbf1e31ff4&benchmark=factorial&view=files&file=mir.mir
when pressing on "dashboard" from this view we should keep the same base vs head imo instead of resetting

figtracer:
- pressing enter after choosing base vs head should work too
- unified vs split should be kept when switching files
- shouldn't let compare the same base vs head mby just a small nit

mattsse (approved):

figtracer (approved):

PR #2061, opened 2026-09-04
docs(lint): document storage aliases, inheritance, and naming exceptions

mattsse:
Documents the lint behavior changes from foundry-rs/foundry#16578 (`pascal-case-struct` preserves leading and trailing underscores and honors `mixed_case_exceptions`), foundry-rs/foundry#16582 (`uninitialized-state` attributes writes through local storage pointers to the state variables they may reference), foundry-rs/foundry#16603 (`missing-events-arithmetic` collects state variables and entry points across the inheritance chain with overrides resolved), foundry-rs/foundry#16607 (`arbitrary-send-erc20` analyzes modifier bodies and treats modifier parameters as safe when every invocation site passes a safe expression), and foundry-rs/foundry#16608 (`ecrecover` tracks struct fields such as `sig.s`).

The `mixed_case_exceptions` entry in the linter config reference now mentions the PascalCase check as well.

### DIFF
--- a/src/pages/config/reference/linter.mdx
+++ b/src/pages/config/reference/linter.mdx
@@ -51,7 +51,7 @@ List of files or patterns to ignore when running the linter. This is a comma-sep
 - Default: `["ERC", "URI"]`
 - Environment: `FOUNDRY_LINT_MIXED_CASE_EXCEPTIONS`
 
-Configurable patterns that should be excluded when performing `mixedCase` lint checks. Defaults to `["ERC", "URI"]` to allow common names like `rescueERC20`, `ERC721TokenReceiver`, or `tokenURI`.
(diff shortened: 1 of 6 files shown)


mablr (approved):

mattsse:
Post-merge documentation follow-up: `mixed_case_exceptions` is still documented here as defaulting to only `["ERC", "URI"]`, while Foundry's `LintSpecificConfig::default()` contains ten entries: `ERC, URI, ID, URL, API, JSON, XML, HTML, HTTP, HTTPS`. Please update both the Default field and the explanatory sentence in a follow-up. I also included this in the open Foundry #16578 review because its Rust doc comment has the same stale list.

PR #2066, opened 2026-09-06
docs(config): document symbolic defaults

figtracer:
Documents `symbolic.check_invariant_frontiers`, introduced by foundry-rs/foundry#16643, including its default, environment variable, dependency on imported invariant frontiers, concrete replay behavior, and an example configuration. It also populates the default configuration's `[symbolic]` block from Foundry's complete `SymbolicConfig` defaults instead of displaying only the new key.

AI assistance: Codex assisted with this documentation under Gustavo Figueiredo's direction.

### DIFF
--- a/src/pages/config/reference/default-config.mdx
+++ b/src/pages/config/reference/default-config.mdx
@@ -771,10 +771,136 @@ show_edge_coverage = false
 # =============================================================================
 
 [symbolic]
+# Whether symbolic tests are enabled
+# Default: false
+enabled = false
+
+# Whether fuzz tests should be symbolically concretized into fuzz corpus seeds
+# Default: false
+seed_corpus = false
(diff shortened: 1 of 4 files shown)


src/pages/config/reference/default-config.mdx
```diff
@@ -766,6 +766,16 @@ corpus_min_size = 0
 # Default: false
 show_edge_coverage = false
 
+# =============================================================================
```
DaniPopes, line 769: is this page out of date or is it intentionally listing only certain configs?

figtracer: out of date, we intentionally never added defaults here until now will do in a follow-up

mattsse (approved):

PR #2072, opened 2026-09-08
fix(perf): render large artifact diffs without blocking

DaniPopes:
LibString's runtime disassembly previously took 47.5 seconds to highlight in Chrome: roughly 36 seconds of Myers diff computation plus expensive word-level decorations. Moving that work to a worker alone did not fix the CPU cost.

This PR uses patience line matching in a cancellable computation worker, then passes the patch through the library's parser into its virtualized `CodeView`. Full original files remain available for expanding context. Patience can produce coarser changes than minimal Myers matching; independent patch-application tests verify exact reconstruction, including the real LibString pair.

Syntax highlighting stays in the library's worker pool. Intraline decorations are disabled because Shiki's decoration splitting is quadratic on these generated files; syntax colors and added/deleted line colors remain. Pierre is pinned to 1.4.1, which includes the large-hunk stack-overflow fix upstream, so the local dependency patch has been removed. The upgrade also includes worker initialization and virtualization/highlighter lifecycle fixes; its new editing APIs are not needed for this read-only viewer.

Five-second computation limit; failures remain failed for the page session. No visibility/revisit retries. Navigating away cancels pending work. Content-based file identities isolate compiler/commit highlighting, and bounded caches reuse completed diffs across visits. Theme and split/unified changes reuse existing results. No backend, credentials, or database changes.

## Chrome measurements

Same pinned real artifacts, fresh browser contexts. Updated code uses a local production build against the public production API; localhost asset delivery and warm CDN/API responses mean these are not a guarantee of deployed end-to-end latency.

| Runtime disassembly | Production before | Updated local build | Revisit |
| --- | --- | --- | --- |
| LibString | 47.54 s | 0.90–1.06 s | 15 ms |
| Nitro | 2.02–2.04 s | 0.53–0.61 s | 15 ms |
| Governor | 2.49–2.63 s | 0.56–0.62 s | 15–16 ms |

Times measure navigation to highlighted output, not plain-text placeholders. Final LibString computation-worker roundtrip: 93–94 ms; largest main-thread task: 84–86 ms instead of 35.9 seconds. Revisit makes no fetches and starts no workers. A slower API sample took 1.87 seconds total.

Rechecked on Pierre 1.4.1: LibString 0.94–1.04 seconds, 93–96 ms computation-worker roundtrip, 84 ms largest main-thread task, and 15 ms revisit. No meaningful speed change from the upgrade alone. All 134 unit tests and 10 browser tests pass locally. The cold-load highlighting regression now uses the requested five-second rendering budget, rather than three seconds; it still requires syntax-highlighted output and does not retry failed rendering. Both GitHub CI jobs pass on 314ded63. Vercel remains blocked on fork-deployment approval; the deployed preview has not yet been verified.

## Validation

- `vp run check`; 134 perf unit tests; 10 browser tests; production perf build.
- 40,000-line repetitive assembly browser regression requires highlighted output within five seconds, including cold worker/grammar startup.
- Exact patch reconstruction tests: empty files, newline variants, patch markers, randomized repetitive edits, large files, and real LibString artifacts.
- Manual Chrome checks: context expansion, virtualized scrolling, split/unified, and theme changes; no console errors.
- Self-review and four-angle simplify review completed. README updated to match the architecture.

The integration uses the [Diffs documented patch and virtualization APIs](https://diffs.com/docs). For comparison, [GitLab documents server-generated, persisted/cached patches and size limits](https://docs.gitlab.com/development/merge_request_concepts/diffs/); worker offloading alone is not a latency strategy.

AI assistance: implementation and testing with Codex.

LibString runtime disassembly, highlighted and virtualized


decofe (approved):

DaniPopes:
@SocketSecurity ignore npm/@pierre/diffs@1.4.1

PR #2076, opened 2026-09-09
docs: clarify EIP-8037 gas accounting

mattsse:
Add a gas-accounting guide explaining regular/execution gas, state gas, the reservoir, ordinary refunds versus state refills, and the distinction between frame measurements, charged receipt gas, block accounting, and required gas limits. Complete the `lastFrameGas` reference with `gasStateUsed` and explicit field boundaries, add an asserted negative-state-delta example in a standalone Amsterdam project, and clarify snapshot limitations, isolation, network activation, and RPC support.

Companion API documentation updates are in foundry-rs/foundry#16772, foundry-rs/forge-std#910, and alloy-rs/alloy#4195.

AI assistance: Codex authored the documentation, Solidity example, and PR description.

### DIFF
--- a/sidebar/forge.ts
+++ b/sidebar/forge.ts
@@ -10,7 +10,7 @@ export const forgeDocs: SidebarItem[] = [
   { text: "Scripting", link: "/forge/scripting" },
   { text: "Debugging", link: "/forge/debugging" },
   { text: "Gas Tracking", link: "/forge/gas-tracking" },
-  { text: "Gas Accounting", link: "/forge/gas-accounting" },
+  { text: "Gas Accounting (EIP-8037)", link: "/forge/gas-accounting" },
   { text: "Formatting", link: "/forge/formatting" },
   { text: "Linting", link: "/forge/linting" },
 ];
--- a/sidebar/forge.ts
+++ b/sidebar/forge.ts
@@ -10,6 +10,7 @@ export const forgeDocs: SidebarItem[] = [
   { text: "Scripting", link: "/forge/scripting" },
   { text: "Debugging", link: "/forge/debugging" },
   { text: "Gas Tracking", link: "/forge/gas-tracking" },
+  { text: "Gas Accounting", link: "/forge/gas-accounting" },
   { text: "Formatting", link: "/forge/formatting" },
   { text: "Linting", link: "/forge/linting" },
 ];
--- a/src/pages/forge/gas-accounting.mdx
+++ b/src/pages/forge/gas-accounting.mdx
@@ -1,0 +1,92 @@
+---
+description: Understand execution gas, state gas, refunds, receipts, and gas limits with EIP-8037.
+---
+
+## Gas accounting
+
+Use a gas **measurement** to understand an execution, a transaction **receipt** to find the charged gas, and a gas **estimate** to choose the limit for a new transaction. These numbers answer different questions, even when they happen to be equal.
+
+[EIP-8037: State Creation Gas Cost Increase](https://eips.ethereum.org/EIPS/eip-8037) separates gas into two dimensions. This page explains that model and how Foundry exposes it. The EIP and its [tracing API proposal](https://github.com/ethereum/execution-apis/pull/852) are evolving; select tools and a node that implement the rules of the network and block you are testing.
+
+### Execution gas and state gas
+
+| Term | What it pays for | What it excludes |
+| --- | --- | --- |
+| **Regular gas / execution gas** | The same dimension under two names: computation, memory expansion, state access, existing-state updates, and transaction intrinsic costs. | State creation charges assigned to the EIP-8037 state dimension; blob gas. |
+| **State gas** | State creation, such as a new storage slot, a new account, or deployed code. Reported net usage already deducts charges undone by state refills or rollback. | The accompanying regular execution costs; blob gas. |
+| **Combined transaction gas limit** (`tx.gas`) | The single gas budget you supply for intrinsic costs and both execution and state gas. | Blob gas, which is accounted for separately. |
+| **Receipt gas used** (`gasUsed`) | The transaction's combined charged gas after ordinary refunds and the applicable calldata floor. | Unused gas, blob gas, and separate network-specific fees. |
+
+A state-changing opcode can incur **both** regular and state gas. State gas does not mean “all gas spent by storage opcodes.” Before EIP-8037, those operations still cost gas, but their costs use the ordinary single-dimensional schedule.
+
+### How the reservoir works
+
+You still supply **one transaction gas limit**. After subtracting intrinsic gas, the EVM allocates a regular gas allowance, subject to the protocol's regular gas cap. Any excess becomes the **state gas reservoir**.
+
+| Charge or operation | Effect |
+| --- | --- |
+| Regular execution charge | Deducts from regular gas only; cannot use the reservoir. |
+| State creation charge | Uses the reservoir first; if it runs out, the remainder draws from regular gas. |
+| State refill | Returns a state charge when the associated creation is undone. It restores the appropriate pools under EIP-8037's refill rules. |
+| Child call | Receives the state reservoir in full. The 63/64 forwarding rule applies to regular gas only. |
+| `GAS` / Solidity `gasleft()` | Reports regular gas only, excluding the reservoir. |
+
+An empty reservoir does **not** mean state gas is disabled: state charges can still be paid from regular gas. Conversely, a large reservoir cannot fund ordinary execution after regular gas runs out.
+
+A nested call's Solidity `{gas: amount}` argument controls its regular gas allowance. It does not impose a combined execution-plus-state budget because the reservoir is passed separately. A `gasleft()` delta can miss state charges paid by the reservoir or include state charges spilling into regular gas. State refills can even make a later `gasleft()` value larger, so unsigned subtraction can revert.
+
+See the EIP's [transaction accounting](https://eips.ethereum.org/EIPS/eip-8037#transaction-level-gas-accounting-reservoir-model) and [call-frame rules](https://eips.ethereum.org/EIPS/eip-8037#gas-accounting-for-halts-and-reverts).
+
+### Two meanings of “refund”
+
+**Ordinary gas refunds** accumulate in the EVM refund counter and are settled at the transaction boundary. [EIP-3529](https://eips.ethereum.org/EIPS/eip-3529) caps the applied refund at one fifth of gas spent; the [EIP-7623 calldata floor](https://eips.ethereum.org/EIPS/eip-7623) can further limit the reduction in charged gas. A refund does not replenish gas available to execute instructions.
+
+**State gas refills** reverse state creation charges during execution, for example when a slot that was zero at transaction start is set and then cleared. They can replenish gas available during execution and are already deducted from net state gas. They are not the ordinary capped refund counter. A reverted or exceptionally halted frame reports zero net state gas, even if it temporarily needed state gas before failing; its regular execution still costs gas.
+
+A successful nested frame can have a **negative state gas delta** when it clears state created by an earlier frame in the same transaction. This is why `Vm.Gas.gasStateUsed` is signed. Transaction-level state gas is nonnegative. Use signed arithmetic for frame deltas and avoid subtracting state refills twice.
+
+### Which value should you use?
+
+| Your question | Value or API | Important boundary |
+| --- | --- | --- |
+| How much regular gas did the last call or creation use? | `vm.lastFrameGas().gasTotalUsed` | Excludes EIP-8037 state gas; includes nested execution. Isolation adds transaction intrinsic gas and the regular calldata floor. |
+| How much net state gas did that frame use? | `vm.lastFrameGas().gasStateUsed` | Already net of state refills; can be negative. Zero without EIP-8037 or after frame rollback. |
+| What is the net sum of those frame components? | `int256(uint256(g.gasTotalUsed)) + int256(g.gasStateUsed)` | A signed measurement, not a sufficient gas limit or necessarily receipt gas. |
+| What ordinary refund did that frame accumulate? | `vm.lastFrameGas().gasRefunded` | Frame counter without isolation; finalized refund for an isolated transaction. Excludes state refills. |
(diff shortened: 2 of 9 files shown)


mattsse (commented):
The measurement / receipt / estimate framing, the decision table, and the `lastCallGas` vs `lastFrameGas` comparison are the parts of this doc set a user will actually come back to; the comparison matches the behaviour in `testdata/paris/cheats/LastCallGas.t.sol`. Suggestions below: one factual fix (the `executionGasUsed` / `regularGasUsed` sentence is backwards), a page structure change (lead with the decision table, compress the EIP mechanics), naming and placement nits, and a replacement example that actually produces a negative state delta, which I ran under Amsterdam with this PR's forge-std. The example needs the snippets project to build for Amsterdam, see that comment.

src/pages/forge/gas-accounting.mdx
```diff
@@ -0,0 +1,92 @@
...
+| RPC tracing on any network | Depends on both the block's rules and the node's tracer support. Optional state fields may be absent; an unsupported `stateGasTracer` can return an RPC error. An absent field means “not reported,” not measured zero. |
+
+When compiling with Solidity 0.8.36, Amsterdam is experimental and also requires `--experimental` (or `experimental = true` in `foundry.toml`).
+
+Pin the [EVM version](/config/reference/solidity-compiler#evm_version), fork block, optimizer settings, and isolation mode when comparing measurements. Use a Foundry version supporting the selected fork and a current forge-std `Vm.Gas` interface containing `gasStateUsed`. For a live network, consult that network's activation schedule rather than assuming Amsterdam rules from a chain name or from Alloy's ability to decode the fields.
+
+RPC field names can also depend on the node/proposal revision. Current Alloy types use `executionGasUsed`; older revisions may call that dimension `regularGasUsed`. Confirm the response schema when choosing your client version.
```
mattsse: This is inverted. Both commits of execution-apis#852 (2 and 6 July) use `regularGasUsed` in the call tracer, the opcode tracer, and `stateGasTracer`; `executionGasUsed` never appears there. `executionGasUsed` was introduced by alloy#4113 on 11 August and revm-inspectors#476 emits it, so today the spec says one name and alloy/reth say the other. A reader following this paragraph against a node that implements the proposal as written gets `None` for every breakdown field. I suggested a serde alias on the three alloy fields in alloy#4195; until that lands, say what is actually true:

```suggestion
RPC field names differ between implementations. The [execution-apis proposal](https://github.com/ethereum/execution-apis/pull/852) names the execution dimension `regularGasUsed`; Alloy and reth use `executionGasUsed`. Check what your node returns before relying on the breakdown fields.
```

src/pages/forge/gas-accounting.mdx
```diff
@@ -0,0 +1,92 @@
...
+
+**State gas refills** reverse state creation charges during execution, for example when a slot that was zero at transaction start is set and then cleared. They can replenish gas available during execution and are already deducted from net state gas. They are not the ordinary capped refund counter. A reverted or exceptionally halted frame reports zero net state gas, even if it temporarily needed state gas before failing; its regular execution still costs gas.
+
+A successful nested frame can have a **negative state gas delta** when it clears state created by an earlier frame in the same transaction. This is why `Vm.Gas.gasStateUsed` is signed. Transaction-level state gas is nonnegative. Use signed arithmetic for frame deltas and avoid subtracting state refills twice.
+
+### Which value should you use?
```
mattsse: This table is the most useful thing in all four PRs. Move it directly under the intro paragraph so it is the first thing a reader sees; the model sections can follow for readers who want the why. As ordered now, a Foundry user has to read the reservoir mechanics and two meanings of refund before reaching the one section that answers their question.

src/pages/forge/gas-accounting.mdx
```diff
@@ -0,0 +1,92 @@
...
+| State refill | Returns a state charge when the associated creation is undone. It restores the appropriate pools under EIP-8037's refill rules. |
+| Child call | Receives the state reservoir in full. The 63/64 forwarding rule applies to regular gas only. |
+| `GAS` / Solidity `gasleft()` | Reports regular gas only, excluding the reservoir. |
+
+An empty reservoir does **not** mean state gas is disabled: state charges can still be paid from regular gas. Conversely, a large reservoir cannot fund ordinary execution after regular gas runs out.
+
+A nested call's Solidity `{gas: amount}` argument controls its regular gas allowance. It does not impose a combined execution-plus-state budget because the reservoir is passed separately. A `gasleft()` delta can miss state charges paid by the reservoir or include state charges spilling into regular gas. State refills can even make a later `gasleft()` value larger, so unsigned subtraction can revert.
+
+See the EIP's [transaction accounting](https://eips.ethereum.org/EIPS/eip-8037#transaction-level-gas-accounting-reservoir-model) and [call-frame rules](https://eips.ethereum.org/EIPS/eip-8037#gas-accounting-for-halts-and-reverts).
```
mattsse: About half the page re-explains the EIP. A Foundry user needs three things: two dimensions exist, which cheatcode returns which, and the decision table. The reservoir mechanics can be two paragraphs plus the EIP anchors. Same for "Transaction gas versus block gas" below, which belongs next to the header and receipt types, i.e. alloy's `state_gas` module doc, rather than in a Forge guide.

```suggestion
### How the reservoir works

You still supply **one transaction gas limit**. After intrinsic gas, the EVM gives the transaction a regular gas allowance (capped by the protocol's per-transaction limit) and puts any excess into a **state gas reservoir**. Regular execution spends regular gas only. State creation spends the reservoir first and spills into regular gas once it is empty; a state **refill** (a slot created and cleared in the same transaction, or a rolled-back creation) credits the pools back in reverse order. Child calls receive the reservoir in full, and the 63/64 rule applies to regular gas only.

`gasleft()` reports regular gas only. A `gasleft()` delta can therefore miss state gas paid from the reservoir, and a refill can make a later `gasleft()` **larger**, which reverts under checked subtraction. See the EIP's [reservoir model](https://eips.ethereum.org/EIPS/eip-8037#transaction-level-gas-accounting-reservoir-model) and [call-frame rules](https://eips.ethereum.org/EIPS/eip-8037#gas-accounting-for-halts-and-reverts) for the full mechanics.
```

src/pages/forge/gas-accounting.mdx
```diff
@@ -0,0 +1,92 @@
...
+
+Use a gas **measurement** to understand an execution, a transaction **receipt** to find the charged gas, and a gas **estimate** to choose the limit for a new transaction. These numbers answer different questions, even when they happen to be equal.
+
+[EIP-8037: State Creation Gas Cost Increase](https://eips.ethereum.org/EIPS/eip-8037) separates gas into two dimensions. This page explains that model and how Foundry exposes it. The EIP and its [tracing API proposal](https://github.com/ethereum/execution-apis/pull/852) are evolving; select tools and a node that implement the rules of the network and block you are testing.
+
+### Execution gas and state gas
+
+| Term | What it pays for | What it excludes |
+| --- | --- | --- |
+| **Regular gas / execution gas** | The same dimension under two names: computation, memory expansion, state access, existing-state updates, and transaction intrinsic costs. | State creation charges assigned to the EIP-8037 state dimension; blob gas. |
```
mattsse: Two names for one dimension get introduced here and then alternate through all four PRs ("regular (execution) gas", "execution gas, also called regular gas", "the same dimension under two names"). Pick one per repo and alias it exactly once. For Foundry I would stay with *regular gas*, which matches `gasleft()` and the `Vm.Gas` field names, and note the EIP's term here:

```suggestion
| **Regular gas** (the EIP's *execution gas*) | Computation, memory expansion, state access, existing-state updates, and transaction intrinsic costs. | State creation charges assigned to the EIP-8037 state dimension; blob gas. |
```

src/pages/forge/gas-tracking.mdx
```diff
@@ -6,6 +6,8 @@ description: Track and compare gas usage with Forge snapshots and reports.
 
 Forge tracks gas usage to help optimize contracts and catch regressions.
 
+For EIP-8037, distinguish regular (execution) gas, state gas, and charged receipt gas. A scalar gas report or snapshot is not a transaction gas-limit estimate. See [gas accounting](/forge/gas-accounting) for what each value includes and [`lastFrameGas`](/reference/cheatcodes/last-frame-gas) for separate frame measurements.
```
mattsse: Placed above "Gas reports", this reads as a warning before the reader knows what a gas report is. A one-line pointer is enough on this page:

```suggestion
Under EIP-8037 (Amsterdam) gas has two dimensions, execution and state. See [gas accounting](/forge/gas-accounting) for how reports, snapshots, and [`lastFrameGas`](/reference/cheatcodes/last-frame-gas) relate to it.
```

sidebar/forge.ts
```diff
@@ -10,6 +10,7 @@ export const forgeDocs: SidebarItem[] = [
   { text: "Scripting", link: "/forge/scripting" },
   { text: "Debugging", link: "/forge/debugging" },
   { text: "Gas Tracking", link: "/forge/gas-tracking" },
+  { text: "Gas Accounting", link: "/forge/gas-accounting" },
```
mattsse: "Gas Tracking" directly above "Gas Accounting" gives no hint which one to open. Naming the EIP does; the alternative is folding this page into Gas Tracking as a section.

```suggestion
  { text: "Gas Accounting (EIP-8037)", link: "/forge/gas-accounting" },
```

src/pages/reference/cheatcodes/gas-snapshots.mdx
```diff
@@ -32,6 +32,14 @@ These cheatcodes allow you to capture gas usage in your tests. Snapshots are wri
 | `snapshotGasLastFrame` | Record gas usage of the last external call or contract creation |
 | `snapshotGasLastCall` | Deprecated call-only variant; use `snapshotGasLastFrame` |
 
+### EIP-8037 gas accounting
+
+A scalar snapshot is not a guaranteed total of regular and state gas. `snapshotGasLastFrame` and the deprecated `snapshotGasLastCall` use gas consumed from the regular counter: that can include state gas spilled into regular gas, but excludes state gas paid from the reservoir. For isolated frames with zero net state gas, they use the isolated transaction's receipt gas instead. They therefore need not equal `lastFrameGas().gasTotalUsed`.
```
mattsse: The clause "for isolated frames with zero net state gas, they use the isolated transaction's receipt gas instead" faithfully describes `stack.rs`, but what it describes is a semantics switch. Measured with forge 1.8.1-nightly, `evm_version = "amsterdam"`, isolation on: a call that creates a slot snapshots `28618` (= `gasTotalUsed`, pre-refund, all state gas dropped); a call that clears it snapshots `22885` (= `28606 − 5721`, post-refund receipt gas). I raised this on foundry#16772 with the suggestion that isolated snapshots always record `tx_gas_used()`. If that lands, this paragraph becomes:

```suggestion
`snapshotGasLastFrame` and the deprecated `snapshotGasLastCall` record one number. Without isolation it is the regular gas the frame spent, which excludes state gas paid from the reservoir; with isolation it is the isolated transaction's receipt gas. Neither is `gasTotalUsed + gasStateUsed`; use [`lastFrameGas`](/reference/cheatcodes/last-frame-gas) for the components.
```

If the behaviour stays as is, this needs a two-row table (state created / not created) rather than a subordinate clause; readers will not extract the switch from the current sentence.

src/snippets/projects/cheatcodes/test/LastFrameGas.t.sol
```diff
@@ -27,6 +27,24 @@ contract LastFrameGasTest is Test {
...
+        int256 netStateGasUsed = int256(gas.gasStateUsed);
+        int256 netMeasuredGas = int256(regularGasUsed) + netStateGasUsed;
+
+        // These are measured components, not a transaction gas-limit estimate.
+        emit log_named_uint("Regular gas (excludes state gas)", regularGasUsed);
+        emit log_named_int("State gas (after state refills)", netStateGasUsed);
+        emit log_named_int("Net sum of measured components", netMeasuredGas);
+        emit log_named_int("Ordinary refund (excludes state refills)", int256(gas.gasRefunded));
+    }
+    // [!endregion components]
```
mattsse: The current example logs four numbers after one call and never shows the case the surrounding prose spends three paragraphs on: a negative `gasStateUsed`. It also has no assertion, so it cannot catch a regression. This replacement produces the negative delta and teaches the isolation point at the same time.

Verified with forge 1.8.1-nightly (2026-09-04), `evm_version = "amsterdam"`, `experimental = true`, and this PR's forge-std: the create frame reports `gasStateUsed = 97920`, the clearing frame `-97920` with `gasTotalUsed = 514`, net `-97406`. With isolation left on, the same sequence reports `0` for the second frame because each call is its own transaction and the slot's original value is then 42, which is worth one sentence in the prose above the include.

Prerequisite: `src/snippets/projects/cheatcodes/foundry.toml` sets no `evm_version`, so the project builds for Osaka today and `gasStateUsed` is always `0` there, for the existing example too. It needs `evm_version = "amsterdam"` and `experimental = true` (solc 0.8.36 refuses Amsterdam without it; I hit that error first). Worth checking that the other snippets in that project still compile under Amsterdam.

```suggestion
    // [!region components]
    /// forge-config: default.isolate = false
    function testNestedFrameCanRefillStateGas() public {
        GasTarget target = new GasTarget();
        target.setValue(42); // creates a slot that was zero at transaction start
        assertGt(vm.lastFrameGas().gasStateUsed, 0);

        target.setValue(0); // clears it again: this frame refills state gas
        Vm.Gas memory gas = vm.lastFrameGas();
        assertLt(gas.gasStateUsed, 0);

        // Widen before adding: the sum is a signed measurement, not a gas limit estimate.
        int256 netMeasuredGas = int256(uint256(gas.gasTotalUsed)) + int256(gas.gasStateUsed);
        emit log_named_int("Net sum of measured components", netMeasuredGas);
    }
    // [!endregion components]
```

mattsse:
Moved the API decision table first, shortened the reservoir explanation, standardized the Foundry terminology, and corrected the proposal’s `regularGasUsed` spelling. The example now asserts a negative state delta with isolation disabled for that test. It has a standalone Amsterdam project with Solidity 0.8.36 and experimental mode; this avoids changing the compiler for existing snippets pinned to older Solidity. The snapshot page explicitly tabulates the current isolation-dependent accounting instead of changing runtime behavior.

AI assistance: Codex authored this update and reply.

mablr (approved):

PR #2079, opened 2026-09-10
docs: clarify expectEmit library call depth

mablr:
Document how registering `expectEmit` expectations in public or external library helpers adds a `delegatecall` frame and affects when expectations are checked. Recommend internal helpers or registering expectations directly in the test before the target call.

Prepared with AI assistance.


figtracer (approved):

anilhelvaci:
This is great!

PR #2092, opened 2026-09-16
fix(perf): improve source viewer rendering

DaniPopes:
Base Solidity highlighting on a local copy of Hardhat's TextMate grammar, with direct improvements for escaped strings, line continuations, nested inheritance expressions, and literals in function modifiers. Keep the upstream attribution, MIT license, and local changes documented alongside the grammar. Use a separate upstream Yul grammar and GitHub light/dark themes for files and diffs.

Make the directory tree easier to scan with compact rows, folder/file icons, and folders sorted first. Selecting a file expands its ancestors without closing directories the user opened; folder state also survives compiler changes. Follows #2091.

Built with Codex assistance.

Before: production source viewer

Solidity: upstream grammar, light theme

Solidity: upstream grammar, dark theme

Yul: dedicated upstream grammar


decofe (approved):

figtracer:
could we show the separate base/head headers only for split diffs where both sides have different contents? they’re misleading in unified mode, and when the contents are identical or only head is present they make a single pane look like two.

PR #2098, opened 2026-09-18
chore(ci): add secure runner across workflows

decofe:
Add the pinned secure-runner action before checkout in all four jobs: checks, Vocs/browser build, perf browser tests, and the scheduled/manual documentation updater. Grant job-scoped OIDC permissions so package downloads from pnpm/Vite+ and the updater's Cargo script run after firewall setup. Preserve frozen installs and existing lockfiles.

Git dependencies, Foundry/solc binaries, and browser downloads are not covered by registry package checks.

Validation: actionlint passed (with existing Depot labels configured where needed), git diff --check passed, and a YAML coverage audit confirmed the first-step ordering, OIDC permissions, preserved triggers/matrices, and intended exclusions. Offline zizmor introduced no new findings relative to the base branch; existing findings remain. GitHub CI and workflows triggered only by schedules/releases/Dependabot still require runtime validation.

The action's documented fork-PR fallback skips package-policy enforcement when GitHub withholds OIDC. Existing caches are retained and are not rescanned.

Prepared with AI assistance.

Prompted by: @grandizzy


.github/workflows/test.yml
```diff
@@ -13,8 +13,11 @@ jobs:
     name: checks
     runs-on: ubuntu-latest
     permissions:
+      id-token: write
       contents: read
     steps:
+      - name: Secure runner
+        uses: tempoxyz/gh-actions/actions/secure-runner@ce5deead22108ccc6aafba0ca47d2a9e6d1a40e8
```
mablr, line 20: `vercel.json:155` still bootstraps Vite+ and runs `vp install --frozen-lockfile` on Vercel, independently of these GitHub jobs (the PR also has a separate Vercel deployment check). The secure-runner step here cannot protect that install.

grandizzy: good point, will look into hardening in follow-up, maybe by downloading and comparing sha, would that work?

grandizzy (approved):

